PR #10, opened 2024-02-13
Add Advanced SIMD three different (for PMULL and PMULL2)

pennyannn:
This PR adds instructions for polynomial multiplication including PMULL and PMULL2


shigoel (approved):
LGTM!

PR #20, opened 2024-02-27
LSR (alias of UBFM) in Data Processing Immediate.

nebeid:
### Description:
Add LSR instruction which is an alias of UBFM and is part of Bitfield instructions of Data Processing - Immediate (DPI)
https://developer.arm.com/documentation/ddi0602/2023-12/Base-Instructions/LSR--immediate---Logical-Shift-Right--immediate---an-alias-of-UBFM-

### Testing:

`make all` including conformance testing.

#### Call out:
There is a specific rand function to add LSR immediate instructions. However, the conformance test output shows mostly UBFX and UBFIZ and a few LSR and LSL in the output. I think the specific function is not being called.

### License:

### DIFF
--- a/Arm/Insts/DPI/Add_sub_imm.lean
+++ b/Arm/Insts/DPI/Add_sub_imm.lean
@@ -39,7 +39,7 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
 ----------------------------------------------------------------------
 
 /-- Generate random instructions of the DPI.Add_sub_imm class. -/
-def Add_sub_imm_cls.rand : IO (Option (BitVec 32)) := do
+def Add_sub_imm_cls.dpi.rand : IO (Option (BitVec 32)) := do
   let (inst : Add_sub_imm_cls) :=
     { sf    := ← BitVec.rand 1,
       op    := ← BitVec.rand 1,
@@ -54,6 +54,8 @@ def Add_sub_imm_cls.rand : IO (Option (BitVec 32)) := do
       Rd    := ← BitVec.rand 5 (lo := 0) (hi := 30) }
   pure (some (inst.toBitVec32))
 
+def Add_sub_imm_cls.rand : List (IO (Option (BitVec 32))) :=
+  [ Add_sub_imm_cls.dpi.rand ]
 ----------------------------------------------------------------------
 
 end DPI
--- a/Arm/Insts/DPI/Add_sub_imm.lean
+++ b/Arm/Insts/DPI/Add_sub_imm.lean
@@ -39,7 +39,7 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
 ----------------------------------------------------------------------
 
 /-- Generate random instructions of the DPI.Add_sub_imm class. -/
-def Add_sub_imm_cls.dpi.rand : IO (Option (BitVec 32)) := do
+def Add_sub_imm_cls.inst.rand : IO (Option (BitVec 32)) := do
   let (inst : Add_sub_imm_cls) :=
     { sf    := ← BitVec.rand 1,
       op    := ← BitVec.rand 1,
@@ -55,7 +55,7 @@ def Add_sub_imm_cls.dpi.rand : IO (Option (BitVec 32)) := do
   pure (some (inst.toBitVec32))
 
 def Add_sub_imm_cls.rand : List (IO (Option (BitVec 32))) :=
-  [ Add_sub_imm_cls.dpi.rand ]
+  [ Add_sub_imm_cls.inst.rand ]
 ----------------------------------------------------------------------
 
 end DPI
--- a/Arm/Insts/DPI/Bitfield.lean
+++ b/Arm/Insts/DPI/Bitfield.lean
@@ -40,18 +40,12 @@ def exec_bitfield (inst: Bitfield_cls) (s : ArmState) : ArmState :=
 ----------------------------------------------------------------------
 
 /-- Generate random instructions of the DPI.Bitfield class. -/
-partial def bitfield_rand_aux (sf : BitVec 1) (N : BitVec 1) (immr : BitVec 6) (imms : BitVec 6) :
-IO (Option (BitVec 32)) := do
-  if sf == 1 && N != 1 then
-    let N := 0b1#1
-    bitfield_rand_aux sf N immr imms
-  else if sf == 0 && (N != 0 || extractLsb 5 5 immr != 0 || extractLsb 5 5 imms != 0) then
-    let sf := ← BitVec.rand 1
-    let N := sf
-    let immr := sf ++ (← BitVec.rand 5)
-    let imms := sf ++ (← BitVec.rand 5)
-    bitfield_rand_aux sf N immr imms
-  else
+partial def Bitfield_cls.ubfm.rand : IO (Option (BitVec 32)) := do
+  -- Choose assignments based on sf that will not result in illegal instructions
+  let sf := ← BitVec.rand 1
+  let N := sf
+  let immr := sf ++ (← BitVec.rand 5)
+  let imms := sf ++ (← BitVec.rand 5)
   let (inst : Bitfield_cls) :=
     { sf    := sf,
       opc   := ← pure 0b10#2,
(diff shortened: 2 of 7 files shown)


Arm/Insts/DPI/Insts.lean
```diff
@@ -6,9 +6,11 @@ Author(s): Shilpi Goel
...
 import Arm.Insts.DPI.Logical_imm
+import Arm.Insts.DPI.Bitfield
 
 /-- List of functions to generate random instructions of the
 DPI class. -/
 def DPI.rand : List (IO (Option (BitVec 32))) :=
   [ DPI.Add_sub_imm_cls.rand,
-    DPI.Logical_imm_cls.rand ]
+    DPI.Logical_imm_cls.rand,
+    DPI.Bitfield_cls.ubfm.rand ]
```
shigoel, line 16: @nebeid Replacing this with `Bitfield_cls.rand` should deal with the call-out, I think.

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,83 @@
+/-
+Copyright (c) 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Shilpi Goel, Nevine
+-/
+-- For now, support only UBFM (immediate) aliased as LSR (immediate):
```
shigoel, line 6: The `exec_bitfield` function below appears to capture the whole UBFM behavior. Confining it to just one of its aliases seems unnecessarily restrictive.

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,83 @@
...
+import Arm.Decode
+import Arm.Insts.Common
+
+namespace DPI
+
+open Std.BitVec
+
+@[simp]
+def exec_bitfield (inst: Bitfield_cls) (s : ArmState) : ArmState :=
+  -- TODO: match opc
```
shigoel, line 18: I'd suggest matching on `opc` and if it isn't equal to `0b10` (UBFM's), throwing an "unimplemented instruction" error (here's an example: https://github.com/leanprover/LNSym/blob/3fc5b4fc614bc4eb954cf65e60949744b96f15b9/Arm/Insts/DPR/Conditional_select.lean#L32)

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,83 @@
...
+  if (inst.sf == 1 && inst.N != 1) ||
+     (inst.sf == 0 && (inst.N != 0 || immr5 != 0 || imms5 != 0)) then
+    write_err (StateError.Illegal s!"Illegal {inst} encountered!") s
+  else
+    let datasize  := if inst.sf == 1#1 then 64 else 32
+    let wtmask    := decode_bit_masks inst.N inst.imms inst.immr false datasize
+    match wtmask with
+    | none => write_err (StateError.Illegal s!"Illegal {inst} encountered!") s
+    | some (wmask, tmask) =>
+      let src := read_gpr datasize inst.Rn s
```
pennyannn, line 30: We need to use `read_gpr_zr` when register 31 is mapped to the zero register ZR. See:
https://github.com/leanprover/LNSym/blob/main/Arm/State.lean#L317
and 
https://stackoverflow.com/questions/61532867/in-arm64-assembly-code-when-is-register-31-xzr-versus-sp

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,83 @@
...
+      let bot := (BitVec.ror src inst.immr.toNat) &&& wmask
+      let result := bot &&& tmask
+      let s := write_gpr datasize inst.Rd result s
```
pennyannn, line 33: `write_gpr_zr` as well.

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,83 @@
...
+  pure (some (inst.toBitVec32))
+
+partial def Bitfield_cls.ubfm.rand : IO (Option (BitVec 32)) := do
+  let sf := ← BitVec.rand 1
+  let N := ← BitVec.rand 1
+  let immr := ← BitVec.rand 6
+  let imms := ← BitVec.rand 6
+  bitfield_rand_aux sf N immr imms
+
+partial def Bitfield_cls.lsr.rand : IO (Option (BitVec 32)) := do
```
pennyannn, line 71: How about defining a single function called `Bitfield_cls.ubfm.rand` that tests just UBFM? Since it has a bunch of alias'es, it will cover the testing of all of them (LSR variants, LSL variants, UBFIZ, UBFX, UXTB and UXTH), including LSR.

nebeid: That is true, I just wanted to ensure that we're testing LSR since it's the instruction we're targeting. The conditions for it are not that likely to occur imms = sf|11111, so I wanted to have a function to enforce it. However, after I directed the second trials in the else if branch in bitfield_rand_aux, I started seeing more LSR and LSL in the conformance test output. Still with the latest change, I don't see that many, but I think some are assembled as UBFIZ and UBFX.

Arm/Insts/DPI/Add_sub_imm.lean
```diff
@@ -39,7 +39,7 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
 ----------------------------------------------------------------------
 
 /-- Generate random instructions of the DPI.Add_sub_imm class. -/
-def Add_sub_imm_cls.rand : IO (Option (BitVec 32)) := do
+def Add_sub_imm_cls.dpi.rand : IO (Option (BitVec 32)) := do
```
pennyannn, line 42: DPI is one level above Add_sub_imm, how about calling it `Add_sub_imm.addsub.rand`?

Arm/Insts/DPI/Bitfield.lean
```diff
@@ -0,0 +1,85 @@
...
+import Arm.Insts.Common
+
+namespace DPI
+
+open Std.BitVec
+
+@[simp]
+def exec_bitfield (inst: Bitfield_cls) (s : ArmState) : ArmState :=
+  if inst.opc != 0b10#2 then
+    write_err (StateError.Illegal s!"Illegal {inst} encountered!") s
```
pennyannn, line 19: `write_err (StateError.Unimplemented s!"Unsupported {inst} encountered!") s`
See: https://github.com/leanprover/LNSym/blob/main/Arm/State.lean#L76

Arm/Insts/DPI/Bitfield.lean
pennyannn: Thanks for the explanation. I was wondering what you meant in the PR description. I think it is a valid concern. The problem is that the logic in `bitfield_rand_aux` is a bit too complex. Do you think if it would be better to have two separate functions (instead of using `bitfield_rand_aux`), one for `ubfm` and one for `lsr`, just to make the logic simpler? And add a comment for the `lsr` function that explains why it is treated special.

nebeid: I thought to do that at the beginning, but then did it this way to avoid repeated checks and code, but I can go back to separating them as I think now that the checks for invalid instructions will be only in the `ubfm` function. The only repeated code will be to create `inst`. Is it preferable to separate them?

pennyannn: I would say yes. @shigoel ?

nebeid: Done. Thank you, Yan, for the advice. It is more readable now and it made me notice that the LSR generation wasn't purely for LSR because of `immr` being generates as 6 random bits instead of making the most significant match `sf`. Now it is only generating LSR instructions as desired.

Arm/Insts/DPI/Bitfield.lean
shigoel: @nebeid I'll second Yan here!

Arm/Insts/DPI/Bitfield.lean
shigoel: Thanks, @nebeid and @pennyannn, and sorry for being late on this!

The `UBFM` and `LSR` random generation functions look very clear now, thanks. However, this does open us up to defining separate generation functions for each instruction alias. Sure, we only care about `LSR` right now, but we may very well want `LSL`, `UXTH`, etc. to be tested too in the future. That solution makes me uncomfortable.

If we increase the number of random tests, we won't need to do the above. Also, we don't have conformance testing running in CI yet, but we will at some point this year --- the issue of not hitting certain values randomly won't be as important because of the volume of tests we'd be running.

For now, we could leave this as is, but I will revisit this later.

Arm/Insts/DPI/Add_sub_imm.lean
shigoel: @nebeid @pennyannn I'd suggest `Add_sub_imm.inst.rand`, given that `Add_sub_imm_cls.rand` is just a list of this one function.

Arm/Insts/DPI/Logical_imm.lean
```diff
@@ -61,7 +61,7 @@ def exec_logical_imm (inst : Logical_imm_cls) (s : ArmState) : ArmState :=
 ----------------------------------------------------------------------
 
 /-- Generate random instructions of the DPI.Logical_imm class. -/
-partial def Logical_imm_cls.rand : IO (Option (BitVec 32)) := do
+partial def Logical_imm_cls.dpi.rand : IO (Option (BitVec 32)) := do
```
shigoel, line 64: Same suggestion as for `Add_sub_imm` and for the same reason: `.dpi` -> `.inst`

nebeid: Updated. Sorry I missed some comments in the last commit.

shigoel (approved):
LGTM! Thanks, @nebeid and @pennyannn!

PR #25, opened 2024-03-05
Add shift related instructions

pennyannn:
### Description:

1. Added [advanced SIMD scalar shift by immediate](https://developer.arm.com/documentation/ddi0602/2023-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en) instructions: SHL, SSHR and USHR, SSRA and USRA, SRSHR and URSHR, SRSRA and URSRA)
2. Added [advanced SIMD (vector) shift by immediate ](https://developer.arm.com/documentation/ddi0602/2023-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en) instructions: SHL, SSHR and USHR, SSRA and USRA, SRSHR and URSHR, SRSRA and URSRA)
3. Added [data processing two source](https://developer.arm.com/documentation/ddi0602/2023-12/Index-by-Encoding/Data-Processing----Register?lang=en) class for variable shifts (LSLV, LSRV, ARSV, RORV 32-, 64-bit)

### Testing:

The `make all` succeeds and conformance testing runs successfully on Graviton2 and Graviton3.

### License:

### DIFF
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -512,7 +512,7 @@ macro_rules
          | _ => $rhsElse:term) => do
   let mut result := rhsElse
   let x ← `(discr)
-  for p in ps, rhs in rhss do
+  for p in ps.reverse, rhs in rhss.reverse do
     let test ← genBVPatMatchTest x p
     let rhs ← declBVPatVars x p rhs
     result ← `(if $test then $rhs else $result)
--- a/Arm/Decode.lean
+++ b/Arm/Decode.lean
@@ -78,6 +78,8 @@ def decode_data_proc_reg (i : BitVec 32) : Option ArmInst :=
     DPR (Conditional_select {sf, op, S, Rm, cond, op2, Rn, Rd})
   | [sf:1, 1, S:1, 11010110, opcode2:5, opcode:6, Rn:5, Rd:5] =>
     DPR (Data_processing_one_source {sf, S, opcode2, opcode, Rn, Rd})
+  | [sf:1, 0, S:1, 11010110, Rm:5, opcode:6, Rn:5, Rd:5] =>
+    DPR (Data_processing_two_source {sf, S, Rm, opcode, Rn, Rd})
   | [sf:1, opc:2, 01010, shift:2, N:1, Rm:5, imm6:6, Rn:5, Rd:5] =>
     DPR (Logical_shifted_reg {sf, opc, shift, N, Rm, imm6, Rn, Rd})
   | _ => none
@@ -104,6 +106,11 @@ def decode_data_proc_sfp (i : BitVec 32) : Option ArmInst :=
     DPSFP (Advanced_simd_permute {Q, size, Rm, opcode, Rn, Rd})
   | [0, Q:1, op:1, 0111100000, a:1, b:1, c:1, cmode:4, o2:1, 1, d:1, e:1, f:1, g:1, h:1, Rd:5] =>
     DPSFP (Advanced_simd_modified_immediate {Q, op, a, b, c, cmode, o2, d, e, f, g, h, Rd})
+  -- Note: Advanced SIMD shift by immediate constraint immh != 0000
+  | [0, Q:1, U:1, 011110, immh:4, immb:3, opcode:5, 1, Rn:5, Rd:5] =>
+    DPSFP (Advanced_simd_shift_by_immediate {Q, U, immh, immb, opcode, Rn, Rd})
+  | [01, U:1, 111110, immh:4, immb:3, opcode:5, 1, Rn:5, Rd:5] =>
+    DPSFP (Advanced_simd_scalar_shift_by_immediate {U, immh, immb, opcode, Rn, Rd})
   | [01, op:1, 11110000, imm5:5, 0, imm4:4, 1, Rn:5, Rd:5] =>
     DPSFP (Advanced_simd_scalar_copy {op, imm5, imm4, Rn, Rd})
   | [0, Q:1, 001110, op2:2, 0, Rm:5, 0, len:2, op:1, 00, Rn:5, Rd:5] =>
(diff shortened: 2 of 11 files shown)


Arm/BitVec.lean
shigoel, line 1: Thanks for this fix to `match_bv`, @pennyannn!

Arm/Decode.lean
```diff
@@ -104,6 +106,11 @@ def decode_data_proc_sfp (i : BitVec 32) : Option ArmInst :=
     DPSFP (Advanced_simd_permute {Q, size, Rm, opcode, Rn, Rd})
   | [0, Q:1, op:1, 0111100000, a:1, b:1, c:1, cmode:4, o2:1, 1, d:1, e:1, f:1, g:1, h:1, Rd:5] =>
     DPSFP (Advanced_simd_modified_immediate {Q, op, a, b, c, cmode, o2, d, e, f, g, h, Rd})
+  -- Note: Advanced SIMD shift by immediate constraint immh != 0000
```
shigoel, line 109: Where is this constraint enforced?

Arm/Insts/Common.lean
```diff
@@ -368,4 +368,89 @@ def elem_set (vector : BitVec n) (e : Nat) (size : Nat)
   have h : hi - lo + 1 = size := by simp only [hi, lo]; omega
   BitVec.partInstall hi lo (h ▸ value) vector
 
+----------------------------------------------------------------------
+
+structure ShiftRightInfo where
```
shigoel, line 373: Curious why you defined `ShiftRightInfo` and `ShiftLeftInfo` instead of just one structure, `ShiftInfo`. The fields in `ShiftRightInfo` that aren't relevant to the left-shift can be optional.

Arm/Decode.lean
pennyannn: I added the note because it is not enforced but inferred. Currently an instruction will be matched against `Advanced_simd_modified_immediate` first, if it doesn't match, then we know `immh` can't be `0b0000#4`.

pennyannn: We will need a more capable `match_bv` to be able to supply `immh != 0b0000#4` as a guard.

Arm/Insts/Common.lean
pennyannn: Great idea.

Arm/Decode.lean
shigoel: Ah, I understand! We can, of course, make `match_bv` take a guard if we come across more of these kinds of issues (and care about even more readability for this decoding), but for now, it serves the purpose well enough.

I'd recommend one of two things to resolve this: 
1. Modify the comment for clarity. i.e., essentially paste in what you say in your first comment above.
2. Delete the comment altogether (I found its existence a little confusing).

nebeid (commented):
I will continue reviewing. Thanks, Yan.

Arm/Insts/Common.lean
```diff
@@ -368,4 +368,81 @@ def elem_set (vector : BitVec n) (e : Nat) (size : Nat)
   have h : hi - lo + 1 = size := by simp only [hi, lo]; omega
   BitVec.partInstall hi lo (h ▸ value) vector
 
+----------------------------------------------------------------------
+
+-- Field unsigned, round and accumulate are optional for left shifts
```
nebeid, line 373: In reality, "optional" means:
- "can be used" or 
- "are not used" and that's why they're set to optional in Lean?

Arm/Insts/DPSFP/Advanced_simd_scalar_shift_by_immediate.lean
```diff
@@ -0,0 +1,120 @@
...
+  match inst.U, inst.opcode with
+  | 0b0#1, 0b01010#5 => exec_shl_scalar inst s
+  | 0b0#1, 0b00000#5
+  | 0b0#1, 0b00010#5
+  | 0b0#1, 0b00100#5
+  | 0b0#1, 0b00110#5
+  | 0b1#1, 0b00000#5
+  | 0b1#1, 0b00010#5
+  | 0b1#1, 0b00100#5
+  | 0b1#1, 0b00110#5 => exec_shift_right_scalar inst s
```
nebeid, line 74: It would be helpful, I think, if there is a comment beside each line as to which instruction this is U, opcode combination indicates.

Arm/Insts/DPSFP/Advanced_simd_shift_by_immediate.lean
nebeid, line 1: Alternatively, when looking again, the code here is mostly the same as in the scalar file. This is confirmed by the fact that, in the Architecture Reference Manual (ARM) each instruction in both has its scalar and vector interpretation, it's just the elements that are of different size. Is there a strong reason to not combine them into one processing as in the ARM?

Arm/Insts/Common.lean
pennyannn: Here it means "are not used" in the case of left shifts.

Arm/Insts/DPSFP/Advanced_simd_shift_by_immediate.lean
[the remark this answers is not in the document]
pennyannn: That's a great observation. However, since the Arm manual named this class as `[Advanced SIMD shift by immediate](https://developer.arm.com/documentation/ddi0602/2023-12/Index-by-Encoding/Data-Processing----Scalar-Floating-Point-and-Advanced-SIMD?lang=en#asimdshf)`, I would like to keep it as the same.

Arm/Decode.lean
```diff
@@ -104,6 +106,11 @@ def decode_data_proc_sfp (i : BitVec 32) : Option ArmInst :=
...
+  | [0, Q:1, U:1, 011110, immh:4, immb:3, opcode:5, 1, Rn:5, Rd:5] =>
+    DPSFP (Advanced_simd_shift_by_immediate {Q, U, immh, immb, opcode, Rn, Rd})
```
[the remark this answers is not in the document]
pennyannn: Same as above.

Arm/Insts/DPSFP/Advanced_simd_scalar_shift_by_immediate.lean
pennyannn: Good idea. I will add that.

Arm/Insts/DPSFP/Advanced_simd_shift_by_immediate.lean
pennyannn: The Arm manual divided them into two categories of classes. So we will need two different types for storing the decoded instructions, and two set of functions to model the instructions. The functions are slightly different. However, you are right they have a lot in common, this is why in common.lean I defined a common type called `ShiftInfo` and functions `shift_right_common`, `shift_left_common` that could be used by both classes.

nebeid (approved):

shigoel (approved):
LGTM! Thanks, @nebeid and @pennyannn!

PR #27, opened 2024-03-25
Change the program type in `ArmState` to `Map` 

shigoel:
### Description:

Change the program type to `Map` (courtesy @aqjune-aws).

Before this change, the `program` field in `ArmState` had the following type:
`Store (BitVec 64) (Option (BitVec 32))`
Now its type is:
`Map (BitVec 64) (BitVec 32)`

An immediate fallout is that the precondition for symbolic simulation/code reasoning will now become:
`s0.program = <program_map>`
instead of
`s0.program = <program_map>.find?`

### Testing:

`make all` succeeded.

### License:

### DIFF
--- a/Proofs/MultiInsts.lean
+++ b/Proofs/MultiInsts.lean
@@ -25,31 +25,31 @@ def test_program : program :=
     (0x126514#64 , 0x4ea21c5c#32),      --  mov     v28.16b, v2.16b
     (0x126518#64 , 0x4ea31c7d#32)]      --  mov     v29.16b, v3.16b
 
-theorem one_asm_snippet_sym_helper1 (q0_var : BitVec 128) :
-  zeroExtend 128 (zeroExtend 128 (zeroExtend 128 q0_var ||| zeroExtend 128 q0_var)) = zeroExtend 128 q0_var := by
-  sorry -- auto
-
 theorem one_asm_snippet_sym_helper2 (q0_var : BitVec 128) :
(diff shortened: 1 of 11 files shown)


Proofs/MultiInsts.lean
```diff
@@ -25,31 +25,31 @@ def test_program : program :=
...
+  (h_s0_pc : read_pc s0 = 0x12650c#64)
+  (h_s0_program : s0.program = test_program)
+  (h_s0_ok : read_err s0 = StateError.None)
+  (h_run : s_final = run 4 s0) :
+  read_sfp 128 26#5 s_final = read_sfp 128 0#5 s0 ∧
+  read_err s_final = StateError.None := by
+  -- Prelude
+  simp_all only [state_simp_rules, -h_run]
+  -- Symbolic Simulation
+  sym_n 4 h_s0_program
```
aqjune-aws, line 41: It is nice to see the use of sym_n here! :)

aqjune-aws (approved):
Looks good to me, thanks.

Proofs/MultiInsts.lean
shigoel: Yes, but something is still less than ideal here. E.g., I want to do away with all of these tactics:
https://github.com/leanprover/LNSym/blob/39be0fa911346cb31c6ec78f510cc694e3188dda/Proofs/MultiInsts.lean#L45-L52

But that's for another day perhaps.

PR #30, opened 2024-05-16
Adding and fixing instructions

pennyannn:
### Description:

This PR contains:
1. Adding instruction class -- Move wide (immediate)
2. Adding instruction class -- Hints (for NOP)
3. Fixing a bug in conditional branch instructions, B.cond
4. Fixing a bug in ADRP
5. Fixing a bug in CMP (immediate)
6. Add instruction class -- Load/store register (unscaled immediate) for LDUR and STUR (SIMD&SFP)
7. Some refactoring

### Testing:

Make all and tests on graviton machines succeed.

### License:

### DIFF
--- a/Arm/Insts/DPI/Add_sub_imm.lean
+++ b/Arm/Insts/DPI/Add_sub_imm.lean
@@ -33,7 +33,9 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
     -- State Updates
     let s'            := write_pc ((read_pc s) + 4#64) s
     let s'            := if setflags then write_pstate pstate s' else s'
-    let s'            := write_gpr datasize inst.Rd result s'
+    -- When Rd == 31, the instruction is the alias CMP <Xn|XSP>,#<imm>[,<shift>]
+    -- This instruction writes out result to XZR
+    let s'            := write_gpr_zr datasize inst.Rd result s'
     s'
 
 ----------------------------------------------------------------------
--- a/Arm/Insts/DPI/Add_sub_imm.lean
+++ b/Arm/Insts/DPI/Add_sub_imm.lean
@@ -33,9 +33,9 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
     -- State Updates
     let s'            := write_pc ((read_pc s) + 4#64) s
     let s'            := if setflags then write_pstate pstate s' else s'
-    -- When Rd == 31, the instruction is the alias CMP <Xn|XSP>,#<imm>[,<shift>]
-    -- This instruction writes out result to XZR
-    let s'            := write_gpr_zr datasize inst.Rd result s'
+    let s'            := if inst.Rd == 31#5 ∧ ¬ setflags
+                         then write_gpr datasize inst.Rd result s'
+                         else write_gpr_zr datasize inst.Rd result s'
     s'
 
 ----------------------------------------------------------------------
--- a/Arm/Insts/DPI/PC_rel_addressing.lean
+++ b/Arm/Insts/DPI/PC_rel_addressing.lean
@@ -20,7 +20,7 @@ def exec_pc_rel_addressing (inst : PC_rel_addressing_cls) (s : ArmState) : ArmSt
   let imm := if inst.op = 0#1 then
                 signExtend 64 (inst.immhi ++ inst.immlo) -- ADR
              else
-                signExtend 64 ((inst.immhi ++ inst.immlo) <<< 12) -- ADRP
+                signExtend 64 (inst.immhi ++ inst.immlo ++ (BitVec.zero 12)) -- ADRP
   let result := if inst.op = 0#1 then
                    orig_pc + imm -- ADR
                 else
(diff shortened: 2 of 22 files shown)


Arm/Insts/DPI/Add_sub_imm.lean
```diff
@@ -33,7 +33,9 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
     -- State Updates
     let s'            := write_pc ((read_pc s) + 4#64) s
     let s'            := if setflags then write_pstate pstate s' else s'
-    let s'            := write_gpr datasize inst.Rd result s'
+    -- When Rd == 31, the instruction is the alias CMP <Xn|XSP>,#<imm>[,<shift>]
+    -- This instruction writes out result to XZR
+    let s'            := write_gpr_zr datasize inst.Rd result s'
```
shigoel, line 38: The comment at L36 is a little misleading because the alias situation is different for each of the instructions being modeled here, and not all of them are only when Rd == 31:
1) ADD (imm) is used by the alias MOV (to/from SP): https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/ADD--immediate---Add--immediate--?lang=en
2) ADDS (imm) is used by the alias CMN (imm): https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/ADDS--immediate---Add--immediate---setting-flags-?lang=en
3) SUB (imm) is used by no alias (https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/SUB--immediate---Subtract--immediate--?lang=en)
4) SUBS (imm) is used by the CMP (imm) alias (https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/SUBS--immediate---Subtract--immediate---setting-flags-?lang=en)

Also, these instructions write to the Stack Pointer, which is the default behavior of `write_gpr` when called with register index 31 -- `write_gpr_zr` essentially throws out the result when register index is 31. I think `write_gpr` should be used here instead of `write_gpr_zr`.

Arm/Insts/DPI/PC_rel_addressing.lean
```diff
@@ -20,7 +20,7 @@ def exec_pc_rel_addressing (inst : PC_rel_addressing_cls) (s : ArmState) : ArmSt
   let imm := if inst.op = 0#1 then
                 signExtend 64 (inst.immhi ++ inst.immlo) -- ADR
              else
-                signExtend 64 ((inst.immhi ++ inst.immlo) <<< 12) -- ADRP
+                signExtend 64 (inst.immhi ++ inst.immlo ++ (BitVec.zero 12)) -- ADRP
```
shigoel, line 23: Ugh, sorry about this bug and thanks for fixing it! I sometimes forget that bitvec operations are fixed-width. :)

Arm/Insts/DPI/Add_sub_imm.lean
pennyannn: This change is added for the last case in 4. It says in the document that 

"CMP <Xn|SP>, #< imm >{, < shift >}

is equivalent to

SUBS XZR, <Xn|SP>, #< imm > {, < shift >}"
which writes out to the XZR register.

I didn't realize there are other alias that might collide with this one though.

pennyannn: It seems like ADDS (like SUBS) also uses XZR: https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/ADDS--immediate---Add--immediate---setting-flags-?lang=en

pennyannn: Oh, actually, I think we can just follow the specification:
```
if d == 31 && !setflags then
    SP[] = result;
else
    X[d] = result;
```
I believe in the else case, it would write to XZR when d == 31.

Arm/Insts/DPI/Add_sub_imm.lean
```diff
@@ -33,7 +33,9 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
     -- State Updates
     let s'            := write_pc ((read_pc s) + 4#64) s
     let s'            := if setflags then write_pstate pstate s' else s'
-    let s'            := write_gpr datasize inst.Rd result s'
+    let s'            := if inst.Rd == 31#5 ∧ ¬ setflags
```
shigoel, line 36: Thanks! 👍

Arm/Insts/DPSFP/Advanced_simd_modified_immediate.lean
```diff
@@ -91,46 +94,50 @@ private theorem mul_div_norm_form_lemma  (n m : Nat) (_h1 : 0 < m) (h2 : n ∣ m
 -- Assumes CheckFPAdvSIMDEnabled64();
 def exec_advanced_simd_modified_immediate
   (inst : Advanced_simd_modified_immediate_cls) (s : ArmState) : ArmState :=
-  if inst.cmode == 0b1111#4 && inst.op == 0b0#1 && inst.o2 == 0b1#1 then
-    write_err (StateError.Unimplemented s!"Unsupported {inst} encountered!") s
-  else if inst.cmode == 0b1111#4 && inst.op == 0b1#1 && inst.Q == 0b0#1 || inst.o2 == 0b1#1 then
+  -- All UNALLOCATED cases when inst.o2 = 1
+  if inst.o2 == 0b1#1 ∧ inst.cmode ++ inst.op != 0b11110#5 then
```
shigoel, line 98: Could this check also be done in `decode_immediate_op`, since this function has access to `inst` anyway?

pennyannn: Sure. Let me try.

shigoel (approved):
Thank you so much!

PR #43, opened 2024-07-19
Use BitVec.cast instead of \t

pennyannn:
### Description:

Use of `BitVec.cast` is preferred instead of `\t`. This PR rewrites uses of `\t` to `BitVec.cast`. 

In doing so, it causes the theorem `sha512h_rule_2` in `Proofs/SHA512/Sha512_block_armv8_rules.lean` to fail. This is because extra `BitVec.cast` is introduced into the goal that could not be recognized by `bv_decide`. By simplifying the goal using rule `BitVec.cast_eq`, `BitVec.cast` is removed from the goal and the theorem goes through again.

### Testing:

"make all" and cosimulation on Graviton2 and Graviton3 succeed.

### License:


shigoel (approved):
Thank you so much, @pennyannn! This looks great.

PR #48, opened 2024-07-30
feat: add Proofs/Experiments that collects examples we will use to experiment when building proofs and automation

bollu:
This allows us to clearly see that we're making progress on the examples that are of interest, and also gives us a subfolder to experiment in.


Proofs/Experiments/Abs.lean
```diff
@@ -0,0 +1,34 @@
+/-
+Copyright (c) 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
shigoel, line 2: Copyright 2024, please!

bollu: Fixed!

shigoel (approved):

PR #51, opened 2024-08-01
chore: Make ArmState fields private to ensure that every field access is written in terms of r/w and read/write_mem_bytes

bollu:
This cleans up our definitions, and forces us to correctly the `read_*` and `write_*` methods. This is important to allow us to reduce all (non)interference reasoning into the `r/w` functions.


Tests/AES-GCM/AESV8ProgramTests.lean
```diff
@@ -28,20 +28,20 @@ open BitVec
...
   assumption that programs don't typically start at pc 0x0#64.
-  If a program runs past the last instruction, its ArmState.error will be:
+  If a program runs past the last instruction, read_err its ArmState will be:
 
     StateError.NotFound "No instruction found at PC 0x0000000000000000#64!"
 
   Based on above information, we could do a binary search for the number of steps
-  to run a program and the stopping criterion is that ArmState.pc = 0x0#64 and
-  ArmState.error = StateError.None. This sounds awfully programmable, but for
+  to run a program and the stopping criterion is that ArmState.pc = 0x0#64 anread_err d
```
shigoel, line 43:
```suggestion
  to run a program and the stopping criterion is that ArmState.pc = 0x0#64 and read_err
```

Arm/MemoryProofs.lean
```diff
@@ -5,6 +5,9 @@ Author(s): Shilpi Goel
 -/
 import Arm.SeparateProofs
 import Arm.FromMathlib
+import Arm.State
+import Arm.Separate
+import Arm.SeparateProofs
```
shigoel, line 10: Redundant import? Maybe kill the one on Line 6 instead.

Tests/AES-GCM/AESV8ProgramTests.lean
```diff
@@ -28,20 +28,20 @@ open BitVec
 --   Run test:
 --     1. aws-lc-build/crypto/crypto_test --gtest_filter="AESTest.ABI"
 
-/- 
+/-
   How to get the number of steps to run for a program?
 
-  When a program finishes, ArmState.pc would be set to 0x0#64 and ArmState.error
+  When a program finishes, ArmState.pc would be set to 0x0#64 read_err and ArmState
```
pennyannn, line 34: The sentence reads weird. Should it say:
```
When a program finishes, `ArmState.pc` would be set to 0x0#64 and `read_err ArmState` ...
```

Tests/AES-GCM/AESV8ProgramTests.lean
```diff
@@ -28,20 +28,20 @@ open BitVec
...
   will be `StateError.None`. If a program hasn't reached the last instruction,
   its ArmState.pc will be some number other than 0x0#64. This is based on an
   assumption that programs don't typically start at pc 0x0#64.
-  If a program runs past the last instruction, its ArmState.error will be:
+  If a program runs past the last instruction, read_err its ArmState will be:
```
pennyannn, line 38: Likewise for this line.

Arm/State.lean
```diff
@@ -120,26 +120,29 @@ def PState.zero : PState :=
...
+  /-- PState -/
+  private pstate     : PState
+  /-- Memory: maps 64-bit addresses to bytes -/
+  private mem        : Store (BitVec 64) (BitVec 8)
+  /--
+  Program: maps 64-bit addresses to 32-bit instructions.
+  Note that we have the following assumption baked into our machine model:
+  the program is always disjoint from the rest of the memory.
+  -/
+  program            : Program
```
shigoel, line 138: Out of scope for this PR, but it may be worth exploring if we ought to use `r` for accessing the `program` field as well, which'd solve the special treatment of this field here.

Tests/AES-GCM/AESV8ProgramTests.lean
bollu, line 38:
```suggestion
  If a program runs past the last instruction, (read_err ArmState) will be:
```

Arm/MemoryProofs.lean
bollu: resolved.

Arm/State.lean
shigoel: Thanks!

shigoel (approved):
LGTM! Thanks, @bollu!

Thanks, @pennyannn, for your review!

PR #53, opened 2024-08-01
Added Stack Pointer (SP) Alignment to symbolic simulation tactic

nwetzler:
### Description:

-- Introduced a new predicate that indicates a bit-vector is n-bit
aligned.
-- CheckSPAlignment is now defined in terms of Aligned.
-- introFetchDecodeLemmas now includes an "h_st_sp_aligned" hypothesis
that is maintained after every symbolic simulation step.  This goal is
automatically when applicable lemmas are present (i.e., Aligned over
AddWithCarry).
-- CheckSPAlignment and Aligned return Prop instead of Bool.
-- Added theorems for alignment of bitvector addition
-- SP Alignment case is explicitly addressed.  It would be nice to
   golf this away

### Testing:

Yes.

### License:


nwetzler:
@shigoel Take a look at commit 977786d.  I think I'd prefer to get these lemmas moved somewhere and see if we can golf the `sha512_block_armv8_test_3_sym` proof before merge.  Let's chat about this.

Arm/Insts/Common.lean
```diff
@@ -54,15 +54,32 @@ def ConditionHolds (cond : BitVec 4) (s : ArmState) : Bool :=
   else
     result
 
+/-- Is the bit-vector x a-bit aligned? -/
+def Aligned (x : BitVec n) (a : Nat) : Prop :=
+  match a with
+    | 0 => True
+    | a' + 1 => (extractLsb a' 0 x) = (BitVec.zero _)
```
bollu, line 61:
```suggestion
def Aligned (x : BitVec n) (a : Nat) : Prop :=
  match a with
  | 0 => True
  | a' + 1 => extractLsb a' 0 x = BitVec.zero _
```

Indent `match` guards to be at the same indent level, and remove unnecessary parens

Arm/Insts/Common.lean
```diff
@@ -54,15 +54,32 @@ def ConditionHolds (cond : BitVec 4) (s : ArmState) : Bool :=
...
+
+-- TODO: Is this useful?
+-- @[simp]
+-- theorem Aligned_zero : Aligned x 0 := by simp [Aligned]
```
bollu, line 65:
```suggestion
```

Kill dead code.

Proofs/SHA512/Sha512_block_armv8.lean
```diff
@@ -80,6 +81,25 @@ theorem sha512_program_test_2_sym (s0 s_final : ArmState)
...
+@[state_simp_rules]
+theorem CheckSPAlignment_BitVecAdd_64_4 (st : ArmState) (y : BitVec 64)
+  (x_aligned : CheckSPAlignment st)
+  (y_aligned : Aligned y 4)
+  :
+  Aligned ((r (StateField.GPR 31#5) st) + y) 4 := by
+  simp_all only [CheckSPAlignment, read_gpr, zeroExtend_eq]
+  generalize (r (StateField.GPR 31#5) st) = x at *
+  apply Aligned_BitVecAdd_64_4 x_aligned y_aligned
+
```
bollu, line 102:
```suggestion
@[state_simp_rules]
theorem CheckSPAlignment_BitVecAdd_64_4 (st : ArmState) (y : BitVec 64)
  (x_aligned : CheckSPAlignment st)
  (y_aligned : Aligned y 4)
  :
  Aligned ((r (StateField.GPR 31#5) st) + y) 4 := Aligned_BitVecAdd_64_4 x_aligned y_aligned
```

Since all of these are definitionally equal, we can directly call `Aligned_BitVecAdd_64_4` and use it to solve this goal.

Proofs/SHA512/Sha512_block_armv8.lean
```diff
@@ -80,6 +81,25 @@ theorem sha512_program_test_2_sym (s0 s_final : ArmState)
...
 -- Test 3:
 
+theorem Aligned_BitVecAdd_64_4 {x : BitVec 64} {y : BitVec 64}
+  (x_aligned : Aligned x 4)
+  (y_aligned : Aligned y 4)
+  :
+  Aligned (x + y) 4 := by
+  unfold Aligned at *
+  simp_all
+  bv_decide
```
bollu, line 91:
```suggestion
theorem Aligned_BitVecAdd_64_4 {x : BitVec 64} {y : BitVec 64}
  (x_aligned : Aligned x 4)
  (y_aligned : Aligned y 4)
  :
  Aligned (x + y) 4 := by
  simp_all [Aligned]
  bv_decide
```

Golf by folding the `unfold` into the `simp_all`.

Proofs/SHA512/Sha512_block_armv8.lean
```diff
@@ -102,6 +122,12 @@ theorem sha512_block_armv8_test_3_sym (s0 s_final : ArmState)
   simp_all only [state_simp_rules, -h_run]
   -- Symbolic simulation
   sym_n 4
+  case h_s1_sp_aligned =>
+    apply CheckSPAlignment_BitVecAdd_64_4
+    case x_aligned =>
+      exact h_s0_sp_aligned
+    case y_aligned =>
+      simp_all (config := {decide := true}) only [state_simp_rules, minimal_theory, bitvec_rules]
```
bollu, line 130:
```suggestion
  case h_s1_sp_aligned => apply Aligned_BitVecAdd_64_4 h_s0_sp_aligned (by decide)
```

- As seen previously, `CheckSPAlignment_BitVecAdd_64_4` is definitionally equal to `Aligned_BitVecAdd_64_4`, so we directly use `Aligned_BitVecAdd_64_4` if we like.
    - Since we made the other arguments implicit, we solve `x_aligned` with `h_s0_sp_aligned`. 
    - Since `y_aligned` is a concrete hypothesis about a concrete bitvector (`⊢ Aligned (18446744073709551600#64) 4`), we discharge it with `decide`. Hurray on us for proving that `Aligned` is decidable :)

Arm/Insts/Common.lean
```diff
@@ -54,15 +54,32 @@ def ConditionHolds (cond : BitVec 4) (s : ArmState) : Bool :=
   else
     result
 
+/-- Is the bit-vector x a-bit aligned? -/
```
bollu, line 57:
```suggestion
/-- `Aligned x a` witnesses that the bitvector `x` is `a`-bit aligned. -/
```

Proofs/SHA512/Sha512_block_armv8.lean
bollu: Actually, after reviewing the rest of the file, I think this theorem is un-necessary even, since it's defeq.

Proofs/SHA512/Sha512_block_armv8.lean
nwetzler: Nice catch!

Proofs/SHA512/Sha512_block_armv8.lean
nwetzler: It's likely unnecessary.  I was following a similar pattern Shilpi had from some work I have uncommitted on `popcount32`.

Proofs/SHA512/Sha512_block_armv8.lean
nwetzler: Excellent!  As I mentioned in a comment to Shilpi on this PR, I'd like to understand why we need this case at all.  We should be able to automate application of alignment lemmas.

nwetzler:
> @shigoel Take a look at commit [977786d](https://github.com/leanprover/LNSym/commit/977786d0f4843041cad80d704736c68eb8c2ffd0). I think I'd prefer to get these lemmas moved somewhere and see if we can golf the `sha512_block_armv8_test_3_sym` proof before merge. Let's chat about this.

Commit has changed, but still looking at feedback on the proof above and how to automate the subgoal.

Proofs/SHA512/Sha512_block_armv8.lean
```diff
@@ -80,6 +81,20 @@ theorem sha512_program_test_2_sym (s0 s_final : ArmState)
...
+  Aligned (x + y) 4 := by
+  simp_all [Aligned]
+  bv_decide
+
+@[state_simp_rules]
+theorem CheckSPAlignment_BitVecAdd_64_4 {st : ArmState} {y : BitVec 64}
+  (x_aligned : CheckSPAlignment st)
+  (y_aligned : Aligned y 4)
+  :
+  Aligned ((r (StateField.GPR 31#5) st) + y) 4 := Aligned_BitVecAdd_64_4 x_aligned y_aligned
```
shigoel, line 97: @nwetzler Let's move these into `Proofs/Insts/Common.lean` so that they're available everywhere. But that file doesn't import LeanSAT, so perhaps `bv_decide` can be replaced by `bv_omega` or something, if it is easy enough to do?

nwetzler: Yeah, that was the downside of moving them.  I'll try with `bv_omega`.

shigoel (approved):
Thanks, @nwetzler!

PR #61, opened 2024-08-06
fix: read the full 128 bits of an SFP register in cosimulation

alexkeizer:
### Description:

This fixes a conformance testing error introduced in #51, by reading the full 128 bits of an SFP register instead of only the lower 64

### Testing:

Conformance testing passed

### License:


shigoel (approved):
Thank you so very much, @alexkeizer, for detecting, debugging, and fixing this!

PR #62, opened 2024-08-06
chore: Test a fix for ELFSage's stack overflow [DONTMERGE]

bollu:
chore: use version of elfsage with PR to test build success

    We test with https://github.com/draperlaboratory/ELFSage/pull/11
    to check that our CI passes, in contrast to https://github.com/leanprover/LNSym/pull/59 which fails.


bollu:
@shigoel since https://github.com/draperlaboratory/ELFSage/pull/11 was merged, you should be able to bump https://github.com/leanprover/LNSym/pull/59 with the new version of ELFSage!

PR #64, opened 2024-08-06
chore: docs and tracing for `sym1_i_n` tactic

alexkeizer:
### Description:

We establish some minor documentation and tracing for the `sym1_i_n` tactic.

Furthermore, we add a simple `simp` pre-processing step to `sym1_i_n`, 
so that it can also deal with hypotheses stated using the more idiomatic `read_err`/`read_pc` family of functions.
This pre-processing happens only once, before starting symbolic evaluation, and should thus not have any significant performance impact.

### Testing:

`make all` passes, including conformance testing

### License:

### DIFF
--- a/Tactics/Sym.lean
+++ b/Tactics/Sym.lean
@@ -173,5 +173,8 @@ Where $PC and $STEPS are concrete constants.
 Note that the tactic will search for assumption of *exactly* these names,
 it won't search by def-eq -/
 elab "sym1_i_n" i:num n:num program:ident : tactic => do
+  Lean.Elab.Tactic.evalTactic (← `(tactic|
+    simp (config := {failIfUnchanged := false}) only [state_simp_rules] at *
+  ))
   for j in List.range n.getNat do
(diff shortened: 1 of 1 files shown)


Tactics/Sym.lean
```diff
@@ -152,10 +155,26 @@ def sym1 (curr_state_number : Nat) (_hProgram : Lean.Ident) : Lean.Elab.Tactic.T
...
+
+/-- `sym1_i_n i n h_program` will symbolically evaluate a program for `n` steps,
+starting from state `i`, where `h_program` is an assumption of the form:
+`s{i}.program = someConcreteProgam`.
+
+The context is assumed to contain hypotheses
+```
+h_s{i}_err : r StateField.ERR s{i} = .None
+h_s{i}_pc  : r StateField.PC  s{i} = $PC
+h_run      : sf = run $STEPS s0
```
shigoel, line 170: @alexkeizer Not an action item right now, but we ought to expect the hypothesis that the stack pointer is aligned too in the near future.

PR #66, opened 2024-08-08
Program proofs using assertional verification methods

shigoel:
### Description:

Experiments re. assertional verification methods

### Testing:

`make all` succeeds.

### License:


Arm/BitVec.lean
```diff
@@ -270,14 +270,17 @@ protected def rand (n : Nat) (lo := 0) (hi := 2^n - 1) : IO (BitVec n) := do
...
   rotateRight x r
 
 /-- Return the `i`-th least significant bit (or `0` if `i >= w`) of
     the `n`-bit bitvector `x`. -/
 @[bitvec_rules]
 abbrev lsb (x : BitVec n) (i : Nat) : BitVec 1 :=
-  BitVec.ofBool (getLsb x i)
+  -- BitVec.ofBool (getLsb x i)
+  have h : (i - i + 1) = 1 := by omega
+  h ▸ BitVec.extractLsb i i x
```
bollu, line 283:
```suggestion
  (BitVec.extractLsb i i x).cast (by omega
```

`BitVec.cast` behaves better.

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+    (0x4005dc#64, 0x4a000020#32), --  eor w0, w1, w0
+    (0x4005e0#64, 0xd65f03c0#32)] --  ret
+
+/-- `Abs` structure: Field `w0` refers to the 32-bit input to the program that
+  is expected to be in register `w0` in the initial state, and `state` refers
+  to the Arm machine state.
+
+  We club the input and the state together so that we can express the
+  post-condition of the program only in terms of this structure `Abs`; see
+  `abs_post`. -/
```
bollu, line 33:
```suggestion
/--
`Abs` structure: Field `w0` refers to the 32-bit input to the program that
is expected to be in register `w0` in the initial state, and `state` refers
to the Arm machine state.

We club the input and the state together so that we can express the
post-condition of the program only in terms of this structure `Abs`; see
`abs_post`.
-/
```

The lean4 style seems to be to move the comments into a separate line when the comment is multiline. This is useful, because life is too short to maintain comment indentation :)

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+structure Abs where
+  w0      : BitVec 32
+  state   : ArmState
+
+/-- State Machine for the `Abs` program: the `next` function only steps the
+  `state` and leaves `w0` unmodified. -/
```
bollu, line 39:
```suggestion
/--
State Machine for the `Abs` program: the `next` function only steps the
`state` and leaves `w0` unmodified.
-/
```

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+def spec (x : BitVec 32) : BitVec 32 :=
+  -- We prefer the current definition as opposed to:
+  -- BitVec.ofNat 32 x.toInt.natAbs
+  -- because the above has functions like `toInt` that do not play well with
+  -- bitblasting/LeanSAT.
+  let msb := BitVec.extractLsb 31 31 x
+  if msb == 0#1 then
+    x
+  else
+    (0#32 - x)
```
bollu, line 63: This is mildly unforunate --- can you request that we add signed compare to the BitVec API?

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+
+/-- Function identifying the cutpoints of the program. -/
+def abs_cut (a : Abs) : Prop :=
+  read_pc a.state = 0x4005d0#64 -- First instruction
+  ∨
+  read_pc a.state = 0x4005e0#64 -- Last instruction
+
+/-- Function that attaches assertions at the cutpoints of this program. -/
+def abs_assert (a : Abs) : Prop :=
+  if (read_pc a.state = 0x4005d0#64) then
```
bollu, line 84:
```suggestion
  if read_pc a.state = 0x4005d0#64 then
```

drop unnecessary parens

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+    abs_pre a
+  else if (read_pc a.state = 0x4005e0#64) then
```
bollu, line 86:
```suggestion
  else if read_pc a.state = 0x4005e0#64 then
```

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,383 @@
...
+             h_af_a4, h_a4_w0,
+             minimal_theory, bitvec_rules]
+  done
+
+-------------------------------------------------------------------------------
+
+theorem partial_correctness :
+  PartialCorrectness Abs := by
+  refine Correctness.partial_correctness_from_verification_conditions
+    ?v1 ?v2 ?v3 ?v4
```
bollu, line 338:
```suggestion
  apply Correctness.partial_correctness_from_verification_conditions
```

I find using `apply` to be clearer.

bollu (commented):
LGTM :)

Arm/BitVec.lean
shigoel: Oh yes, of course. Thx!

Proofs/Experiments/AbsVCG.lean
shigoel: Can't take the Lisper out of me.

Proofs/Experiments/AbsVCG.lean
shigoel: Actually, BitVec does have [`slt`](https://github.com/leanprover/lean4/blob/30cf3bb3bfdac33a7ab385b25e0d262e0e3d8599/src/Init/Data/BitVec/Basic.lean#L363) and [`sle`](https://github.com/leanprover/lean4/blob/30cf3bb3bfdac33a7ab385b25e0d262e0e3d8599/src/Init/Data/BitVec/Basic.lean#L370) that correspond to SMTLIB functions `bvslt` and `bvsle` (respectively). So we could rewrite this `spec` to use `slt`.

Arm/BitVec.lean
```diff
@@ -270,14 +270,15 @@ protected def rand (n : Nat) (lo := 0) (hi := 2^n - 1) : IO (BitVec n) := do
...
+@[bitvec_rules]
 abbrev ror (x : BitVec n) (r : Nat) : BitVec n :=
   rotateRight x r
 
 /-- Return the `i`-th least significant bit (or `0` if `i >= w`) of
     the `n`-bit bitvector `x`. -/
 @[bitvec_rules]
 abbrev lsb (x : BitVec n) (i : Nat) : BitVec 1 :=
-  BitVec.ofBool (getLsb x i)
+  (BitVec.extractLsb i i x).cast (by omega)
```
alexkeizer, line 281: If you use `extractLsb'` (which is phrased in terms of `lo` and length, rather than `hi-lo`) you don't need the cast

Proofs/Experiments/AbsVCG.lean
```diff
@@ -0,0 +1,333 @@
...
+def spec (x : BitVec 32) : BitVec 32 :=
+  -- We prefer the current definition as opposed to:
+  -- BitVec.ofNat 32 x.toInt.natAbs
+  -- because the above has functions like `toInt` that do not play well with
+  -- bitblasting/LeanSAT.
+  let msb := BitVec.extractLsb 31 31 x
+  if msb == 0#1 then
+    x
+  else
+    (0#32 - x)
```
alexkeizer, line 44: It should be doable to prove this definition equal to `x.toInt.natAbs` by rewriting along `[toInt_eq_msb_cond]`(https://leanprover-community.github.io/mathlib4_docs/Init/Data/BitVec/Lemmas.html#BitVec.toInt_eq_msb_cond).

That said, for this experiment it makes total sense not to bother with that and just use the more convenient definition you have here

Arm/BitVec.lean
shigoel: Indeed. Unfortunately, `extractLsb'` doesn't seem to be supported by LeanSAT. Here's an illustration:

```
theorem extractLsb_and_Lsb' (x : BitVec 32) :
  BitVec.extractLsb 31 31 x = BitVec.extractLsb' 31 1 x := by
  /-
  -- bv_decide
  Spurious counterexample:
  The prover found a potential counter example, consider the following assignment:
  x = 0xffffffff#32
  BitVec.extractLsb' 31 1 x = 0x0#1
  -/
  simp only [Nat.reduceSub, Nat.reduceAdd, BitVec.extractLsb]
```

shigoel: And upon leaving out `(x : BitVec 32)` above, we get a different error message from LeanSAT:
`None of the hypotheses are in the supported BitVec fragment....`

shigoel: Reported to Henrik on Zulip.

PR #68, opened 2024-08-08
Adding Popcount32 proof

nwetzler:
### Description:

We have successfully symbolically simulated a 27-instruction Arm program compiled from the C function `popcount32` from [Sean Anderson’s Bit Twiddling Hacks](https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel).

### Testing:

Yes

### License:


Proofs/Popcount32.lean
```diff
@@ -0,0 +1,129 @@
...
+       clear h_step_11 h_step_12 h_step_13 h_step_14;
+       clear h_step_15 h_step_16 h_step_17 h_step_18;
+       clear h_step_19 h_step_20;
+       clear h_step_21 h_step_22 h_step_23 h_step_24;
+       clear h_step_25 h_step_26)
+  -- Final Steps
+  unfold run at h_run
+  subst s_final
+  unfold popcount32_spec
+  sorry
```
shigoel, line 127: @nwetzler It'd be good to say why this proof is skipped right now. I assume it's because the instruction effects aren't aggregated automatically to obtain the program effects, right?

nwetzler: Ahh, sorry, I forgot to save in VSCode before I pushed.  Meant to take this theorem out until it was ready.

Arm/Insts/Common.lean
```diff
@@ -109,6 +109,14 @@ theorem CheckSPAligment_of_write_mem_bytes :
...
   simp_all only [CheckSPAlignment, state_simp_rules, minimal_theory, bitvec_rules]
 
+@[state_simp_rules]
+theorem CheckSPAlignment_AddWithCarry_64_4 (st : ArmState) (y : BitVec 64) (carry_in : BitVec 1)
+  (x_aligned : CheckSPAlignment st)
+  (y_carry_in_aligned : Aligned (BitVec.add (extractLsb 3 0 y) (zeroExtend 4 carry_in)) 4)
+  : Aligned (AddWithCarry (r (StateField.GPR 31#5) st) y carry_in).fst 4 := by
+  simp_all only [CheckSPAlignment, read_gpr, zeroExtend_eq, Nat.sub_zero, add_eq,
+    Aligned_AddWithCarry_64_4]
+
```
bollu, line 119:
```suggestion
  simp_all [CheckSPAlignment, read_gpr, zeroExtend_eq, Nat.sub_zero, add_eq,
    Aligned_AddWithCarry_64_4]

```

Since it's a terminal simp (i.e., a simp which closes a goal), it's safe to not need the `only`.

Proofs/Popcount32.lean
```diff
@@ -0,0 +1,129 @@
...
+   (0x400610#64 , 0xb9000fe0#32), -- str w0, [sp, #12]
+   (0x400614#64 , 0xb9400fe0#32), -- ldr w0, [sp, #12]
+   (0x400618#64 , 0x910043ff#32), -- add sp, sp, #0x10
+   (0x40061c#64 , 0xd65f03c0#32)] -- ret
+
+
+#genStepTheorems popcount32_program namePrefix:="popcount32_" thmType:="fetch"
+
+-- #guard_msgs in
+-- #check popcount32_fetch_0x4005b4
```
bollu, line 72:
```suggestion
```

delete dead code?

Proofs/Popcount32.lean
```diff
@@ -0,0 +1,129 @@
...
+  match i with
+  | 0 => BitVec.zero 32
+  | i' + 1 =>
+    let bit_idx := BitVec.getLsb x i'
+    ((BitVec.zeroExtend 32 (BitVec.ofBool bit_idx)) + (popcount32_spec_rec i' x))
+
+def popcount32_spec (x : BitVec 32) : BitVec 32 :=
+  popcount32_spec_rec 32 x
+
+-- #eval popcount32_spec 15#32
```
bollu, line 36:
```suggestion
```

delete dead code?

Proofs/Popcount32.lean
```diff
@@ -0,0 +1,129 @@
...
+/-
+Source: https://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel
+
+int popcount_32 (unsigned int v) {
+  v = v - ((v >> 1) & 0x55555555);
+  v = (v & 0x33333333) + ((v >> 2) & 0x33333333);
+  v = ((v + (v >> 4) & 0xF0F0F0F) * 0x1010101) >> 24;
+  return(v);
+}
+-/
```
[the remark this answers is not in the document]
bollu: I would also keep the code in a comment block in markdown syntax. Unfortunately, if I use the github UI to offer that suggestion, it gets mixed up with its own code block!

bollu (commented):
LGTM!

Arm/Insts/Common.lean
shigoel: Actually, I'd prefer using `only`, even in the terminal case, to keep our proofs robust in case future updates to Lean break us, e.g., maybe by causing some rewrite loops?

Proofs/Popcount32.lean
nwetzler: I think we want to keep this.

bollu: @nwetzler In which case, can we uncomment it with the guard message? Or is that too large?

shigoel (approved):

PR #69, opened 2024-08-09
feat: introduce `sym1_n` tactic as a more intelligent version of `sym1_i_n`

alexkeizer:
### Description:

This refactors the symbolic evaluation using pre-generated theorems to maintain a context structure which tracks the names of the various theorems needed for eval, rather than relying on these theorems to be available with hard-coded names.

In the process, we also define a way to infer the initial context by searching through the local context for variables and hypothesis that are def-eq to what we expect (basically doing `by assumption`). This has two big benefits:
* It allows us to generate more meaningful error messages when the contract is broken, and
* During this search we also infer and assert more information. For example, we assert that the `m` in `sf = run m s0` is a numeric literal and reflect the value (we can use this in future to throw a meaningful error when the user asks to simulate more than `m` steps). Similarly, we reflect the PC (which is also assumed to be a concrete literal) and determine the name of the program; this information will be useful to more intelligently find the right `stepi` lemma to apply in future (in contrast to the current approach which assumes the right lemmas to be present in a simp-set).

Since the search matches up-to defeq, it might impact performance slightly. That said, the search only happens *once* per `sym_n` call, independent of the amount of steps we simulate, so this cost should pale in comparison to the cost of the actual symbolic eval.

I've demonstrated the new tactic works on `gcm_gmult_v8`. In a next PR, I intend to remove `sym_i_n` and replace all uses with `sym_n`.

### Testing:

`make all` passed on the cloud desktop, including conformance testing

### License:

### DIFF
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -203,9 +203,9 @@ def fromLocalContext (state? : Option Name) : MetaM SymContext := do
   }
 where
   findLocalDeclUsernameOfType? (expectedType : Expr) : MetaM (Option Name) := do
-    let some h_run ← findLocalDeclWithType? expectedType
+    let some fvarId ← findLocalDeclWithType? expectedType
       | return none
-    let decl := (← getLCtx).get! h_run
+    let decl := (← getLCtx).get! fvarId
     -- ^^ `findLocalDeclWithType?` only returns `FVarId`s which are present in the local context,
     --    so we can safely pass it to `get!`
     return decl.userName
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -169,10 +169,6 @@ where
       | throwError "Failed to find a local hypothesis of type {expectedType}"
     return name
 
-/-- `mkName` is a simple helper for turning a string into an unqualified name,
-as used for local variables and hypotheses -/
-private def mkName : String → Name :=
-  Lean.Name.mkStr Lean.Name.anonymous
 
 def default (curr_state_number : Nat) : SymContext :=
   let s := s!"s{curr_state_number}"
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -79,8 +79,9 @@ private def reflectNatLiteral (e : Expr) : MetaM Nat := do
   if e.hasFVar then
     throwError "Expected a ground term, but {e} has free variables"
 
-  let e ← instantiateMVars e
-  let some x := e.nat? | throwError "Expected a numeric literal, found:\n\t{e}"
+  let e' ← reduce (← instantiateMVars e)
+  let some x := e'.rawNatLit? | throwError "Expected a numeric literal, found:\n\t{e'}\nwhich was obtained by reducing:\n\t{e}"
+  -- ^^ The previous reduction will have reduced a canonical-form nat literal into a raw literal
   return x
 
 /-- For a concrete width `w`,
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -179,8 +179,8 @@ def default (curr_state_number : Nat) : SymContext :=
     h_run     := .mkSimple s!"h_run"
     h_program := .mkSimple s!"h_{s}_program"
     h_pc      := .mkSimple s!"h_{s}_pc"
-    h_err     := .mkSimple s!"h_{s}_err"
-    h_sp      := .mkSimple s!"h_{s}_sp"
+    h_err     := some <| .mkSimple s!"h_{s}_err"
+    h_sp      := some <| .mkSimple s!"h_{s}_sp"
     /-
       `runSteps`, `pc` and `program` actually require inspection of the context to determine.
       However, these values are not actually used yet,
@@ -206,10 +206,10 @@ def nextState (c : SymContext) (nextPc? : Option (BitVec 64) := none) : SymConte
   {
     state     := .mkSimple s
     h_run     := c.h_run
-    h_program := mkName s!"h_{s}_program"
+    h_program := .mkSimple s!"h_{s}_program"
     h_pc      := .mkSimple s!"h_{s}_pc"
-    h_err     := .mkSimple s!"h_{s}_err"
-    h_sp      := .mkSimple s!"h_{s}_sp"
+    h_err     := some <| .mkSimple s!"h_{s}_err"
+    h_sp      := some <| .mkSimple s!"h_{s}_sp"
     runSteps  := c.runSteps - 1
     program   := c.program
     pc        := nextPc?.getD (c.pc + 4#64)
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -1,0 +1,225 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Alex Keizer
+-/
+import Lean
+import Arm.Exec
+
+/-!
+This files defines the `SymContext` structure,
+which is collects
+-/
+
+open Lean Meta
+open BitVec
+
+/--
+
+-/
+structure SymContext where
+  /-- `state` and `finalState` are local variable of type `ArmState` -/
+  state : Name
+  /-- `runSteps` is the number of steps that we can *maximally* simulate,
+  because of the way it occurs in `h_run`.
+  Note that `runSteps` is a meta-level natural number, reflecting the fact that we expect the number
+  of steps in `h_run` to be expressed as a concrete literal -/
+  runSteps : Nat
+  /-- `h_run` is a local hypothesis of the form `finalState = run {runSteps} state` -/
+  h_run : Name
+  /-- `program` is a *constant* which represents the program being evaluated -/
+  program : Name
+  /-- `h_program` is a local hypothesis of the form `state.program = program` -/
+  h_program : Name
+  /-- `pc` is the *concrete* value of the program counter -/
+  pc : BitVec 64
+  /-- `h_pc` is a local hypothesis of the form `r StateField.PC state = pc` -/
+  h_pc  : Name
+  /-- `h_err` is a local hypothesis of the form `r StateField.ERR state = .None` -/
+  h_err : Option Name
+  /-- `h_sp` is a local hypothesis of the form `CheckSPAlignment state` -/
+  h_sp  : Option Name
+
+  /-- `state_prefix` is used together with `curr_state_number`
+  to determine the name of the next state variable that is added by `sym` -/
+  state_prefix      : String := "s"
+  /-- `curr_state_number` is incremented each simulation step,
+  and used together with `curr_state_number`
+  to determine the name of the next state variable that is added by `sym`  -/
+  curr_state_number : Nat := 0
+
+
+namespace SymContext
+
+/-! ## Creating initial contexts -/
+
+/-- Infer `state_prefix` and `curr_state_number` from the `state` name as follows:
+if `state` is `s{i}` for some number `i` and a single character `s`,
+then `s` is the prefix and `i` the number,
+otherwise simple take `{state}_` as the prefix and start counting from zero -/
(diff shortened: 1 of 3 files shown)


Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Alex Keizer
+-/
+import Lean
+import Arm.Exec
+
+/-!
+This files defines the `SymContext` structure,
+which is collects
```
bollu, line 11:
```suggestion
This files defines the `SymContext` structure,
which tracks the various theorems needed for evalulation. This simplifies implementing proof automation, since automation can look for `SymContext` in the hypothesis, rather than hardcoded theorem names.
```

Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+    { ctxt with
+      state_prefix := state ++ "_",
+      curr_state_number := 0 }
+
+
+private def reflectBitVecLiteral (w : Nat) (e : Expr) : MetaM (BitVec w) := do
+  let x ← mkFreshExprMVar (Expr.const ``Nat [])
+  let e' ← mkAppM ``BitVec.ofNat #[toExpr w, x]
+  if !(←isDefEq e e') then
+    throwError "Not def-eq:\n\t{e}\nand\n\t{e'}" -- TODO: error message
```
bollu, line 81:
```suggestion
    throwError "Not def-eq:\n\t{e}\nand\n\t{e'}" -- TODO: error message
```

Can we change the signature of `reflectBitVecLiteral` so we take a piece of `Syntax` so we can `throwErrorAt`?

Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+  let h_pc_type ← mkEq (← mkAppM ``r #[(.const ``StateField.PC []), stateExpr]) pc
+  let h_pc ← findLocalDeclUsernameOfTypeOrError h_pc_type
+
+  -- Unwrap and reflect `pc`
+  let pc ← instantiateMVars pc
+  let pc ←
+    try
+      reflectBitVecLiteral 64 pc
+    catch e =>
+      throwError "{e.toMessageData}\nIn\n\t{h_pc} : {h_pc_type}"
```
bollu, line 154:
```suggestion
      throwError "{e.toMessageData}\nIn\n\t{h_pc} : {h_pc_type}"
```

`throwErrorAt` pls?

Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+  return inferStatePrefixAndNumber {
+    state, h_run, runSteps, program, h_program, pc, h_pc,
+    h_err := none,
+    h_sp := none,
+  }
+where
+  findLocalDeclUsernameOfType? (expectedType : Expr) : MetaM (Option Name) := do
+    let some h_run ← findLocalDeclWithType? expectedType
+      | return none
+    let decl := (← getLCtx).get! h_run
```
bollu, line 165:
```suggestion
    let decl := (← getLCtx).get! h_run
```

the `get!` is suspicious, a comment explaining why this is OK would be nice.

Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+    return decl.userName
+  findLocalDeclUsernameOfTypeOrError (expectedType : Expr) : MetaM Name := do
+    let some name ← findLocalDeclUsernameOfType? expectedType
+      | throwError "Failed to find a local hypothesis of type {expectedType}"
+    return name
+
+/-- `mkName` is a simple helper for turning a string into an unqualified name,
+as used for local variables and hypotheses -/
+private def mkName : String → Name :=
+  Lean.Name.mkStr Lean.Name.anonymous
```
bollu, line 175:
```suggestion
```

This already exists as `Lean.Name.mkSimple`


```lean
abbrev mkSimple (s : String) : Name :=
  .str .anonymous s
```

Tactics/SymContext.lean
```diff
@@ -0,0 +1,230 @@
...
+
+def default (curr_state_number : Nat) : SymContext :=
+  let s := s!"s{curr_state_number}"
+  {
+    state     := mkName s
+    h_run     := mkName s!"h_run"
+    h_program := mkName s!"h_{s}_program"
+    h_pc      := mkName s!"h_{s}_pc"
+    h_err     := mkName s!"h_{s}_err"
+    h_sp      := mkName s!"h_{s}_sp"
```
bollu, line 185:
```suggestion
  let s := s!"s{curr_state_number}"
  {
    state     := .mkSimple s
    h_run     := .mkSimple s!"h_run"
    h_program := .mkSimple s!"h_{s}_program"
    h_pc      := .mkSimple s!"h_{s}_pc"
    h_err     := .mkSimple s!"h_{s}_err"
    h_sp      := .mkSimple s!"h_{s}_sp"
```

bollu (commented):
@alexkeizer and I chatted about this in-person. He corrected my faulty memory that we could `throwErrorAt` for hypotheses. He suggested that we should try to throw a more informative error message regardless. 

My memory now suggests that it maybe possible to build a `MessageData` with the `FVarId`, and it will then correctly interpolate into the nice hover-with-type-information.

Tactics/SymContext.lean
alexkeizer: This is incorrect: `SymContext` tracks `Name`s at meta-time, it would never show up in the local context

Tactics/SymContext.lean
alexkeizer: I don't think we can: the bitvec comes from the local context so we don't have its syntax.

Note that the current setup should throw the error at the tactic invocation, as is common for tactics.
I do try to make the variable in question part of the error message to give that context (and you're right in calling me out, I should make this error better)

Tactics/SymContext.lean
alexkeizer: See previous comment

Tactics/SymContext.lean
alexkeizer: Added a comment

alexkeizer:
I've refactored the error messages to include the highlighting suggested and generally made them a bit more informative.
Also, I've refactored the reflection of Nats to include reduction (of ground terms).

@bollu Could you please have another look and confirm I've addressed your comments

alexkeizer:
> @alexkeizer and I chatted about this in-person. He corrected my faulty memory that we could `throwErrorAt` for hypotheses. He suggested that we should try to throw a more informative error message regardless.
> 
> My memory now suggests that it maybe possible to build a `MessageData` with the `FVarId`, and it will then correctly interpolate into the nice hover-with-type-information.

For future reference: `FVarId` does not have a `ToMessageData` instance, but `Expr` does: which can be used through `Expr.fvar _` (which is how I've implemented it)

Tactics/SymContext.lean
```diff
@@ -0,0 +1,257 @@
...
+  let pc ← withErrorContext h_pc h_pc_type <| reflectBitVecLiteral 64 pc
+
+  return inferStatePrefixAndNumber {
+    state, h_run, runSteps, program, h_program, pc, h_pc,
+    h_err := none,
+    h_sp := none,
+  }
+where
+  findLocalDeclUsernameOfType? (expectedType : Expr) : MetaM (Option Name) := do
+    let some h_run ← findLocalDeclWithType? expectedType
```
shigoel, line 192: Let's use some name other than `h_run` here.

Tactics/SymContext.lean
```diff
@@ -0,0 +1,257 @@
...
+  Note that `runSteps` is a meta-level natural number, reflecting the fact that we expect the number
+  of steps in `h_run` to be expressed as a concrete literal -/
+  runSteps : Nat
+  /-- `h_run` is a local hypothesis of the form `finalState = run {runSteps} state` -/
+  h_run : Name
+  /-- `program` is a *constant* which represents the program being evaluated -/
+  program : Name
+  /-- `h_program` is a local hypothesis of the form `state.program = program` -/
+  h_program : Name
+  /-- `pc` is the *concrete* value of the program counter -/
```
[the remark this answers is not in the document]
alexkeizer: Following the chat we had, I've added the following to `pc`'s docstring:

> Note that for now we only support symbolic evaluation of programs at statically known addresses.
  At some point in the near future, we will want to support addresses of the type `base +/- offset`
  as well, where `base` is an arbitrary variable and `offset` is statically known.
  We could do so by refactoring `pc` to be of type `Bool × BitVec 64`,
  so long as we assume the instruction addresses in a single program will either be all
  statically known, or all offset against the same `base` address,
  and we assume that no overflow happens (i.e., `base - x` can never be equal to `base + y`)

Tactics/SymContext.lean
```diff
@@ -0,0 +1,257 @@
...
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Alex Keizer
+-/
+import Lean
+import Arm.Exec
+
+/-!
+This files defines the `SymContext` structure,
+which is collects
```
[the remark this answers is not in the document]
alexkeizer: Fixed

Tactics/SymContext.lean
```diff
@@ -0,0 +1,257 @@
...
+  let state := ctxt.state.toString
+  let tail := state.toSubstring.drop 1
+
+  if let some curr_state_number := tail.toNat? then
+    { ctxt with
+      state_prefix := (state.get? 0).getD 's' |>.toString,
+      curr_state_number }
+  else
+    { ctxt with
+      state_prefix := state ++ "_",
```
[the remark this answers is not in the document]
alexkeizer: Crucially, the underscore is added in the `else` branch, so when the state variable *isn't* a single character followed by a number, e.g., if the initial state is called `initialState`, or `state`.

Admittedly, I didn't give it that much thought, this would indeed behave badly if there are multiple `sym_n` calls.
I've refactored it to ignore the `state` variables name if it can't be decomposed, and start counting from `s1` (assuming the initial state should've been `s0`) --- the alternative would be some logic that tries to split on the last underscore and checks if the rest is a `Nat` to ensure round-tripping, but that seemed too complex.

Tactics/SymContext.lean
```diff
@@ -0,0 +1,257 @@
...
+  let s := s!"s{curr_state_number}"
+  {
+    state     := .mkSimple s
+    h_run     := .mkSimple s!"h_run"
+    h_program := .mkSimple s!"h_{s}_program"
+    h_pc      := .mkSimple s!"h_{s}_pc"
+    h_err     := some <| .mkSimple s!"h_{s}_err"
+    h_sp      := some <| .mkSimple s!"h_{s}_sp"
+    /-
+      `runSteps`, `pc` and `program` actually require inspection of the context to determine.
```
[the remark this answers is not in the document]
alexkeizer: I've rewrapped `SymContex.lean` to be limited to 80 characters per line (unfortunately I'm not an emacs user, but I prefer to wrap lines manually, at semantically meaningful points anyway, to get nicer diffs).

I think I've seen existing code that is wider: do we only want to limit only comments to 80, but are fine with code that is longer, or is the style to limit everything to 80, and we should wrap that code at some point, too?

alexkeizer:
@shigoel I've addressed your comments, another round of review would be appreciated!
Also did another run of conformance checking, all is still good.

(I did mess up the history, sorry about that, but the diff w.r.t. the comment you last reviewed would be uninformative because of the line-wrapping anyway)

Tactics/SymContext.lean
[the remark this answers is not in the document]
shigoel: Thanks! You're right -- some existing code and comments are longer. 80col is what we're striving for though. While exceptions can be made for code, comments ought to be wrapped at 80col for readability.

BTW, I'm a VSCode newbie, but I found the [ReWrap extension](https://stackoverflow.com/questions/43122175/automatically-hard-wrap-lines-at-column-in-vscode) useful.

shigoel (approved):
Thanks, @alexkeizer!

PR #56, opened 2024-08-02
Implement a decision procedure for memory reads/write via omega

bollu:
This reduces goals about memory to goals about natural numbers, with the hope of allowing the use of `omega` to dispatch such goals.

PR #70, opened 2024-08-09
chore: move memory into a semarate Arm/Memory/ subdirectory

bollu:
This is motivated by more files that will pop up for automation via https://github.com/leanprover/LNSym/pull/56/.

It feels like the memory subsystem is now large enough to warrant its own subfolder.

### License:


shigoel (approved):
Thanks, @bollu!

PR #72, opened 2024-08-09
refactor: rename `Program.min`/`max` to `min?`/`max?` to reflect it returns an option, and introduce `min`, `min!`, `max`, `max!` wrappers around them with the usual meanings

alexkeizer:
### Description:

We rename `Program.min`/`max` to `min?`/`max?` to reflect that these functions return
 an option. We then introduce `min`, `min!`, `max` and `max!` wrappers around them with the usual meanings (see, [`Option.get`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Option/Basic.html#Option.get) and [`Option.get!`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Option/BasicAux.html#Option.get!)), except that `get` also has `by decide` as an auto-tactic value for its proof.

This is particularly nice because we tend to work with concrete programs, in which case `by decide` can easily tell that a minimal/maximal address exists, and we don't have to deal with the `Option`.

See `GCMGmultV8Sym.lean` for an instance of `Program.min`, which now returns a `BitVec` directly (I've not removed the now-unneeded `Option.some.injEq` from the simp to avoid merge conflicts with #69)

### Testing:

`make all` succeeds on ARM

### License:


alexkeizer:
@bollu Review is appreciated!

bollu:
LGTM, but I'd prefer to have it be called `Program.{min,max}Addr{!?}?`

bollu:
Feel invited to rename in a subsequent PR.

shigoel (approved):
Thanks!

PR #73, opened 2024-08-09
feat: add extractLsByte(s) API for memory.

bollu:
This allows us to eventually write functions in the memory model that work byte-at-a-time.
This PR is peeled from https://github.com/leanprover/LNSym/pull/56.


Arm/BitVec.lean
```diff
@@ -813,4 +813,84 @@ macro_rules
...
+    (decide (i ≤ 7) && val.getLsb (n * 8 + i)) := by
+  simp [extractLsByte]
+  rw [Nat.succ_mul]
+  simp only [Nat.add_one_sub_one]
+  simp [Nat.add_sub_cancel_left]
+
+
+/-! ### Least Significant Byte range -/
+
+/-- grab 'n' least significant bytes, starting from index 'base'.
```
shigoel, line 880:
```suggestion
/-- Get `n` least significant bytes of `val`, starting from index `base`.
```

Arm/BitVec.lean
```diff
@@ -813,4 +813,84 @@ macro_rules
...
+-/
+def extractLsBytes (val : BitVec w) (base : Nat) (n : Nat) : BitVec (n * 8) :=
+  match h : n with
+  | 0 => 0#0
+  | x + 1 => val.extractLsb (base * 8 + n * 8 - 1) (base * 8) |>.cast (by omega)
+
+@[simp]
+theorem getLsb_extractLsBytes (val : BitVec w) (base : Nat) (n : Nat) (i : Nat) :
+    (BitVec.extractLsBytes val base n).getLsb i =
+      (decide (0 < n) && (decide (i ≤ base * 8 + (n) * 8 - 1 - base * 8) && val.getLsb (base * 8 + i))) := by
```
shigoel, line 891: Quibble: let's try to wrap long lines at column 80.

Arm/BitVec.lean
```diff
@@ -813,4 +813,84 @@ macro_rules
...
+  val.extractLsb ((base + 1)*8 - 1) (base * 8) |>.cast (by omega)
+
+theorem extractLsByte_def (val : BitVec w₁) (n : Nat) :
+    val.extractLsByte n = (val.extractLsb ((n + 1)*8 - 1) (n * 8) |>.cast (by omega)) := rfl
+
+@[simp]
+theorem getLsb_extractLsByte (val : BitVec w₁) :
+    ((BitVec.extractLsByte val n).getLsb i) =
+    (decide (i ≤ 7) && val.getLsb (n * 8 + i)) := by
+  simp [extractLsByte]
```
shigoel, line 872: Stick to `simp only` in LNSym as far as possible.

Arm/BitVec.lean
bollu: [Mathlib allows column length 100](https://github.com/leanprover-community/mathlib4/blob/055d766c68bd10e70d7adf8fe87f44add3d806fa/Mathlib/Tactic/Linter/Lint.lean#L259C9-L260). Happy to switch to column length 80 if desired.

bollu:

Arm/BitVec.lean
bollu: Fixed.

Arm/BitVec.lean
shigoel: Yes, please. Thanks!

shigoel (approved):

PR #74, opened 2024-08-12
feat: New definitions for the memory model

bollu:
For freedom in experimenting with definitions, we define our own version of `read_mem` and `write_mem` called `read_mem'` and `write_mem`'. These operate directly on the memory, rather than the `ArmState`. We prove their equivalence to the existing definitions (`read_mem_eq_read_mem'`, `write_mem_eq_write_mem'`).

Furthermore, we define equivalences to `BitVec.extractLsByte` to ease reasoning about byte-level manipulation. As an easy corollary, we get `getLsb` theorems on these in order to allow `omega` based reasoning about bit-level values of memory.

We then build lemmas that allow simplification of the proof state given the new `mem_subset'` and `mem_separate'` assumptions. In total, this gives us automation to simplify theorems about memory (non)-interference.

### Testing:

- `make all` succeeded.
- Conformance testing was not run, since the definitions have not changed.

### License:


Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+      simp only [BitVec.add_zero]
+      have hmod : i % 8 = i := Nat.mod_eq_of_lt h₁
+      simp [hmod]
+    · simp only [h₁, decide_False, cond_false]
+      rw [ih]
+      by_cases h₂ : i - 8 < n' * 8
+      · simp only [h₂, decide_True, Bool.true_and, show (i < n' * 8 + 8) by omega]
+        have hi' : ∃ i', i = i' + 8 := by
+          apply Classical.byContradiction
+          intros h
```
alexkeizer, line 739:
```suggestion
          by_contra h
```

`by_contra` is not a mathlib-ism, is it?

Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+        · omega
+        · omega
+      · simp only [h₂, decide_False, Bool.false_and, Bool.false_eq, Bool.and_eq_false_imp,
+          decide_eq_true_eq]
+        intros h₃
+        omega
+      · omega
+
+/--
+A variant of `write_mem` that directly talks about writes to memory, instead of over the entire `ArmState`,
```
alexkeizer, line 762:
```suggestion
A variant of `write_mem` that directly talks about writes to memory, instead of over the entire `ArmState`
```

Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+-/
+def write_mem' (addr : BitVec 64) (val : BitVec 8) (m : Memory) : Memory :=
+  write_store addr val m
+
+theorem write_mem'_of_eq (hix : ix = addr) : write_mem' addr val s ix = val := by
+  simp only [write_mem']
+  subst ix
+  apply store_read_over_write_same
```
alexkeizer, line 770: Should this not be phrased in terms of `read_mem'`, instead of the direct function application?

The fact that memory is implemented as a function seems like an implementation detail we shouldn't expose in theorems

alexkeizer: Idem for all the other theorems that talk about reading from memory in terms of the function application

Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+
+/-- Writing zero bytes does not change memory. -/
+theorem write_mem_bytes'_zero : write_mem_bytes' 0 addr val s = s := rfl
+
+/-- Writing (n + 1) bytes can be described as writing `n` bytes and then recursing to write the rest. -/
+theorem write_mem_bytes'_succ :
+    write_mem_bytes' (n + 1) addr val s =
+    let byte := BitVec.extractLsb 7 0 val
+    let s := write_mem' addr byte s
+    let val_rest := BitVec.zeroExtend (n * 8) (val >>> 8) -- TODO: rewrite this as 'truncate'.
```
alexkeizer, line 807: Is this a "pre-merge" TODO, or a "follow up PR" TODO?

alexkeizer (commented):
In general looks good, I like having more specific intermediate abstractions where they can be useful by themselves.

I do have some comments about naming, and how much of the abstraction we want to expose.

Arm/State.lean
bollu: This is a "follow up" TODO.

Arm/State.lean
bollu: Sorry Alex, it's in batteries :(

Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+  | 0 => 0#0
+  | n' + 1 =>
+    let byte := read_mem' addr s
+    let rest := read_mem_bytes' n' (addr + 1#64) s
+    have h: n' * 8 + 8 = (n' + 1) * 8 := by simp_arith
+    BitVec.cast h (rest ++ byte)
+
+theorem read_mem_bytes_eq_read_mem_bytes' (s : ArmState) :
+    read_mem_bytes n addr s = read_mem_bytes' n addr s.mem := by
+  induction n generalizing addr s
```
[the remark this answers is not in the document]
bollu: Aha, this is in fact in the right order! See that the LHS operates on `s` and the RHS on `s.mem`.

Arm/State.lean
bollu: I'll mark these as `protected` in order to convey the design principle that these are meant to be low-level APIs that builds the higher level API on top of it.

Arm/State.lean
```diff
@@ -662,3 +662,297 @@ theorem write_mem_irrelevant :
   simp [read_mem, write_mem, store_write_irrelevant]
 
 end Memory
+
+section NewMemory
+/!-
```
alexkeizer, line 667:
```suggestion
/-!
```
Oops, my bad

Arm/State.lean
```diff
@@ -662,3 +662,297 @@ theorem write_mem_irrelevant :
...
+theorem read_mem_bytes_eq_read_mem_bytes' (s : ArmState) :
+    read_mem_bytes n addr s = read_mem_bytes' n addr s.mem := by
+  induction n generalizing addr s
+  case zero => simp [read_mem_bytes, read_mem_bytes']
+  case succ n' ih =>
+    simp [read_mem_bytes, read_mem_bytes', read_mem_eq_read_mem', ih]
+
+@[simp]
+theorem read_mem_bytes'_zero_eq : read_mem_bytes' 0 addr s = 0#0 := rfl
+
```
alexkeizer, line 713:
```suggestion
@[simp]
theorem read_mem_bytes'_zero_eq : read_mem_bytes' 0 addr s = 0#0 := 
  rfl

```

Arm/State.lean
```diff
@@ -662,3 +662,297 @@ theorem write_mem_irrelevant :
...
+theorem read_mem_bytes_eq_read_mem_bytes' (s : ArmState) :
+    read_mem_bytes n addr s = read_mem_bytes' n addr s.mem := by
+  induction n generalizing addr s
+  case zero => simp [read_mem_bytes, read_mem_bytes']
+  case succ n' ih =>
+    simp [read_mem_bytes, read_mem_bytes', read_mem_eq_read_mem', ih]
+
+@[simp]
+theorem read_mem_bytes'_zero_eq : read_mem_bytes' 0 addr s = 0#0 := rfl
```
alexkeizer, line 712:
```suggestion
@[simp]
theorem read_mem_bytes'_zero_eq : read_mem_bytes' 0 addr s = 0#0 := 
  rfl
```

Arm/State.lean
bollu: Can't mark them protected since they're in the root namespace. Rather, I'm going to write a docstring calling these out as low-level

bollu:
@shigoel All comments of Alex's are addressed. Review (and merge) appreciated ^^c

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
 
 end Memory
+
+section NewMemory
+/-!
+# New definitions for the memory model
+
+For freedom in experimenting with definitions,
+we define our own version of `read_mem` and `write_mem`,
+called `read_mem'` and `write_mem`'.
```
shigoel, line 672:
```suggestion
called `read_mem'` and `write_mem'`.
```

Arm/State.lean
```diff
@@ -662,3 +662,296 @@ theorem write_mem_irrelevant :
...
+In total, this gives us automation to simplify theorems about memory (non)-interference.
+-/
+
+abbrev Memory := Store (BitVec 64) (BitVec 8)
+
+/--
+A variant of `read_mem` that directly talks about writes to memory, instead of over the entire `ArmState`,
+-/
+def read_mem' (addr : BitVec 64) (m : Memory) : BitVec 8 :=
+  read_store addr m
```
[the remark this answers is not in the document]
shigoel: I'm happy with namespacing in LNSym -- we really should've done that when we started out, tbh.

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
+  rfl
+
+def read_mem_bytes' (n : Nat) (addr : BitVec 64) (s : Memory) : BitVec (n * 8) :=
+  match n with
+  | 0 => 0#0
+  | n' + 1 =>
+    let byte := read_mem' addr s
+    let rest := read_mem_bytes' n' (addr + 1#64) s
+    have h : n' * 8 + 8 = (n' + 1) * 8 := by simp_arith
+    BitVec.cast h (rest ++ byte)
```
shigoel, line 704:
```suggestion
def read_mem_bytes' (n : Nat) (addr : BitVec 64) (m : Memory) : BitVec (n * 8) :=
  match n with
  | 0 => 0#0
  | n' + 1 =>
    let byte := read_mem' addr m
    let rest := read_mem_bytes' n' (addr + 1#64) m
    have h : n' * 8 + 8 = (n' + 1) * 8 := by simp_arith
    BitVec.cast h (rest ++ byte)
```

Let's consistently use `m` for memory and `s` for state to make it easy to immediately spot which definitions we're talking about. Please also check this for lemmas

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
+-/
+def write_mem' (addr : BitVec 64) (val : BitVec 8) (m : Memory) : Memory :=
+  write_store addr val m
+
+/--
+This is a low level theorem.
+Prefer using theorems from `Arm.Memory.Separate` that provide higher level theorems
+in terms of memory (non)-interference.
+-/
+theorem write_mem'_of_eq (hix : ix = addr) : write_mem' addr val s ix = val := by
```
shigoel, line 776:
```suggestion
theorem write_mem'_of_eq (hix : ix = addr) : write_mem' addr val m ix = val := by
```
Please also check in other lemmas

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
+
+/--
+A variant of `read_mem` that directly talks about writes to memory, instead of over the entire `ArmState`
+-/
+def read_mem' (addr : BitVec 64) (m : Memory) : BitVec 8 :=
+  read_store addr m
+
+theorem read_mem_eq_read_mem' : read_mem addr s = read_mem' addr s.mem := rfl
+
+@[simp]
```
shigoel, line 693: Why the `simp` tag? Time to have a memory-specific simp attribute? We typically do not use general-purpose `simp` lemmas in LNSym.

Arm/State.lean
shigoel: Forgive my ignorance: why is `truncate` preferred over `zeroExtend`?

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
+  induction n generalizing base mem ix
+  case zero => simp [write_mem_bytes']
+  case succ n ih =>
+    simp only [write_mem_bytes']
+    rw [ih]
+    · have hix : ix.toNat > base.toNat := by omega
+      obtain hix : ix.toNat ≠ base.toNat := by omega
+      apply write_mem'_of_neq (by apply BitVec.neq_of_toNat_neq hix)
+    · rw [BitVec.toNat_add_eq_toNat_add_toNat (by simp; omega)]
+      simp; omega
```
shigoel, line 855: `simp only...` please, and please check in other spots too.

Arm/State.lean
```diff
@@ -662,3 +662,318 @@ theorem write_mem_irrelevant :
...
+          rw [Nat.add_mul]
+          omega
+        · simp only [hi', decide_False, Bool.false_and, Bool.false_eq]
+          apply BitVec.getLsb_ge
+          rw [Nat.add_mul, Nat.add_mul]
+          omega
+    · simp [hj]
+
+/--
+The byte at location `ix` in memory, such that `base ≤ ix ≤ base + ix` will be the `ix - base` byte of data.
```
shigoel, line 886:
```suggestion
The byte at location `ix` in memory, such that `base ≤ ix ≤ base + ix` will be the `ix - base` byte of `data`.
```

shigoel (commented):
Nice!

Left some comments, @bollu.

Arm/State.lean
bollu: We allow `simp; omega`, since they are both closing tactics. This is [standard in Lean upstream](https://github.com/leanprover/lean4/blob/11be29e68c8dea408f6bc74033d14181d015a526/src/Init/Data/BitVec/Bitblast.lean#L265). I've checked that `simp`'s that are not followed by `omega` are `simp only`.

Arm/State.lean
bollu: Aha, it felt like the correct operation, but I realize that I was thinking about it wrong, and that we are in fact zero extending. Happy to remove the TODO.

Arm/State.lean
bollu: Done.

bollu:
@shigoel @alexkeizer I've migrated the namespaces, and IMO the PR is good to merge.

Arm/State.lean
```diff
@@ -663,13 +669,13 @@ theorem write_mem_irrelevant :
...
 
-section NewMemory
+
 /-!
 # New definitions for the memory model
 
 For freedom in experimenting with definitions,
 we define our own version of `read_mem` and `write_mem`,
-called `read_mem'` and `write_mem`'.
+called `read_mem'` and `write_mem'`.
```
shigoel, line 678:
```suggestion
called `Memory.read` and `Memory.write`.
```

Arm/State.lean
```diff
@@ -663,13 +669,13 @@ theorem write_mem_irrelevant :
...
 These operate directly on the memory, rather than the `ArmState`.
 We prove their equivalence to the existing definitions
 (`read_mem_eq_read_mem'`, `write_mem_eq_write_mem'`).
```
shigoel, line 681: Change these theorems' names `read_mem_eq_read_mem'`...

Arm/State.lean
```diff
@@ -680,52 +686,61 @@ As an easy corollary, we get `getLsb` theorems on these to allow
...
+theorem ArmState.write_mem_eq_mem_write :  (write_mem addr val s).mem = s.mem.write addr val := rfl
 
-@[simp]
-theorem getLsb_read_mem' : (read_mem' addr s).getLsb i = (s addr).getLsb i :=
-  rfl
+namespace Memory
+
+/--
+A variant of `read_mem` that directly talks about writes to memory, instead of over the entire `ArmState`
+-/
```
shigoel, line 706: Should be moved above the definition of `Memory.read` above

Arm/State.lean
shigoel: Yes, but in order to be more robust to the changes in general simp sets upstream, we prefer to use `only` in LNSym as much as possible.

Arm/State.lean
```diff
@@ -662,3 +668,327 @@ theorem write_mem_irrelevant :
...
+    write_mem_bytes n addr val s =
+    { s with mem := s.mem.write_bytes n addr val } := by
+  induction n generalizing addr s
+  case zero => simp [write_mem_bytes, write_bytes_zero]
+  case succ n' ih =>
+    simp [write_mem_bytes, ArmState.read_mem_eq_mem_read, ih,
+      write_mem, write_bytes]
+    rfl
+
+/-- Writing (n + 1) bytes can be described as writing `n` bytes and then recursing to write the rest. -/
```
alexkeizer, line 829: Probably more than 80 chars wide, I feel

Arm/State.lean
```diff
@@ -662,3 +668,327 @@ theorem write_mem_irrelevant :
...
+theorem ArmState.write_mem_eq_mem_write :  (write_mem addr val s).mem = s.mem.write addr val := rfl
+
+namespace Memory
+
+/--
+A variant of `read_mem` that directly talks about writes to memory, instead of over the entire `ArmState`
+-/
+
+@[memory_simp_rules]
+theorem getLsb_read (mem : Memory) : (mem.read addr).getLsb i = (mem addr).getLsb i := rfl
```
alexkeizer, line 709: Why do we need this theorem? It feels like it's getting rid of the abstraction boundary, so it really should only be an internal lemma, but then it's part of a simp-set

alexkeizer:
Thanks for going through the painful effort of putting it all in namespaces, it feels much cleaner to me already!

Arm/State.lean
bollu: @alexkeizer `getLsb` is meant to be simplified away in the internal theory, but I'm happy to un-simp it if you like.

Arm/State.lean
bollu: Done! thanks

Arm/State.lean
bollu: Done :)

Arm/State.lean
```diff
@@ -662,3 +668,327 @@ theorem write_mem_irrelevant :
...
+def Memory.write (addr : BitVec 64) (val : BitVec 8) (m : Memory) : Memory :=
+  write_store addr val m
+
+theorem ArmState.write_mem_eq_mem_write :  (write_mem addr val s).mem = s.mem.write addr val := rfl
+
+namespace Memory
+
+/--
+A variant of `read_mem` that directly talks about writes to memory, instead of over the entire `ArmState`
+-/
```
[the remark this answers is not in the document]
bollu: Done!

Arm/State.lean
bollu: Fixed.

Arm/State.lean
```diff
@@ -662,3 +668,327 @@ theorem write_mem_irrelevant :
...
+Furthermore, we define equivalences to `BitVec.extractLsByte`
+to ease reasoning about byte-level manipulation.
+As an easy corollary, we get `getLsb` theorems on these to allow
+`omega` based reasoning about bit-level values of memory.
+-/
+
+def Memory.read (addr : BitVec 64) (m : Memory) : BitVec 8 :=
+  read_store addr m
+
+theorem ArmState.read_mem_eq_mem_read : read_mem addr s = s.mem.read addr := rfl
```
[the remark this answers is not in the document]
bollu: @alexkeizer I'm going to decide simp sets once I have more experience with the API, hope that's okay

bollu:
Fixed, @shigoel @alexkeizer :)

shigoel (approved):
Thanks!

PR #76, opened 2024-08-12
refactor: generate `stepi` theorems in the relevant program's namespace

alexkeizer:
### Description:

Change `#genStepTheorems` theorem generation so that the generated theorems are put in the namespace of the program they belong to.

This removes the need for the user to pick a prefix, but most importantly, gives these theorems predictable names, so that other automation (i.e., `sym_n`) knows where to find them.

### Testing:

`make all` succeeds, I've not run conformance testing, since I haven't changed any definitions

### License:

### DIFF
--- a/Proofs/SHA512/Sha512DecodeExecLemmas.lean
+++ b/Proofs/SHA512/Sha512DecodeExecLemmas.lean
@@ -4,7 +4,7 @@ import Proofs.SHA512.Sha512Program
 
 -- set_option trace.gen_step.debug.heartBeats true in
 -- set_option trace.gen_step.print_names true in
-set_option maxHeartbeats 2000000 in
+set_option maxHeartbeats 1005000 in
 #genStepTheorems sha512_program thmType:="decodeExec"
 
 /--
--- a/Proofs/SHA512/Sha512DecodeExecLemmas.lean
+++ b/Proofs/SHA512/Sha512DecodeExecLemmas.lean
@@ -4,7 +4,7 @@ import Proofs.SHA512.Sha512Program
 
 -- set_option trace.gen_step.debug.heartBeats true in
 -- set_option trace.gen_step.print_names true in
-set_option maxHeartbeats 1005000 in
+set_option maxHeartbeats 1002500 in
 #genStepTheorems sha512_program thmType:="decodeExec"
 
 /--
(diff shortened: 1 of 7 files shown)


Proofs/SHA512/Sha512DecodeExecLemmas.lean
```diff
@@ -4,11 +4,11 @@ import Proofs.SHA512.Sha512Program
 
 -- set_option trace.gen_step.debug.heartBeats true in
 -- set_option trace.gen_step.print_names true in
-set_option maxHeartbeats 1000000 in
-#genStepTheorems sha512_program namePrefix:="sha512_" thmType:="decodeExec"
+set_option maxHeartbeats 2000000 in
```
shigoel, line 7: Why the increase in `maxHeartbeats`? I know this may be a stopgap measure because we have plans to speed up this autogeneration of lemmas, but this is worth paying attention to.

alexkeizer: I conjecture this is because of my change to use `s!"..."` syntax.
Apparently this introduces a bunch of redundant `... ++ "" ++ ...` terms, which are allocations and thus bump the heartbeats. I'll revert that change and see if it makes a difference (we really should set up that benchmark to see if this actually makes a difference to performance)

```lean
def foo (a b : String) : String := s!"{a}{b}"
#print foo /-
def foo : String → String → String :=
fun a b => toString "" ++ toString a ++ toString "" ++ toString b ++ toString ""
-/
```

alexkeizer: Nope, rolling that back did not fix it. I had a look at some of the other changes I made, but I can't really pinpoint what causes us to need the extra heartbeats.

I did manage to find a tighter limit, we really only need `1002500` (i.e., `2500` more than before). This seems insignificant enough, especially considering I plan to completely rewrite most of the generation anyway. I'm happy to setup a benchmark so that we can compare actual performance, if you're still worried

shigoel: Alright, good enough for now, thanks. Let's wait on that benchmark plan until the new generation framework is in place.

shigoel (approved):
Thanks!

PR #77, opened 2024-08-13
Modify cosim scripts to work on M* Apple machines

shigoel:
### Description:

Cosimulation can now be run on Arm-based Apple machines.

### Testing:

`make all` and conformance testing succeeded on an Apple M3 machine and an Amazon Linux Graviton2 machine.

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #78, opened 2024-08-13
feat: Aarch64 runner

bollu:
If this works, this can be merged after #77 to enable cosim in github CI.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -14,7 +14,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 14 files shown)


shigoel:
Amd64? Do you mean Arm64/Aarch64?

bollu:
@shigoel Feel invited to hit merge, the CI action does the right thing now :)

shigoel (approved):
Thank you!

shigoel:
@bollu Thanks for submitting the issue: https://github.com/leanprover/lean-action/issues/83 
Can we create one of our own that uses `leanprover/lean-action` when the above is resolved?

PR #80, opened 2024-08-13
feat: add simproc to simplify fetch_inst through reflection

alexkeizer:
### Description:

We add a `ProgramInfo` struct, which for now is just a representation of a `Program` as a `HashMap`.
This struct will eventually be stored in the environment, and will cache relevant reflected information about a program.

For now, we use `ProgramInfo` to make a `simproc` that simplifies `fetch_inst` by simply looking up the right instruction in the `ProgramInfo`, and then constructing a proof that this is indeed equal to the original expression.

We most likely don't want to use this simproc in `sym` as-is, but the machinery should be generic enough to be repurposed for stepi-lemma generation.

In the process I've also moved a bunch of stuff from `SymContext` to `Tactics.Common`:
- Everything that was removed from the former is moved to the latter without changes
- The only new code in `Tactics.Common` is `getBitVecValue?`
- The other files contain only new code

I'm happy to split this into seperate refactor and new-code PRs, if that is easier to review.

### Testing:

`make all` succeeds locally; I've not run conformance testing

### License:


Tactics/Common.lean
```diff
@@ -78,3 +78,99 @@ def getStateFieldString? (e : Expr) : MetaM (Option String) := OptionT.run do
...
+-- TODO: should this be upstreamed to core?
+def getBitVecValue? (e : Expr) : MetaM (Option ((n : Nat) × BitVec n)) :=
+  match_expr e with
+    | BitVec.ofFin _ i => OptionT.run do
+      let ⟨n, i⟩ ← getFinValue? i
+      let n' := Nat.log2 n
+      if h : n = 2^n' then
+        return ⟨n', .ofFin (Fin.cast h i)⟩
+      else
+        failure
```
bollu, line 99:
```suggestion
        throwError s!"Expected (BitVec.ofFin (Fin.mk {n} {i})) to have {n} be a power of two."
```

alexkeizer: Note that this is in an `OptionT`, so failure here will return `MetaM.pure none`, this function is not supposed to throw an error if it fails reflection

Tests/Tactics/ReduceFetchInst.lean
```diff
@@ -0,0 +1,13 @@
...
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Alex Keizer
+-/
+import Tactics.Reflect.FetchAndDecode
+import Tests.«AES-GCM».GCMGmultV8Program
+
+/-! ## Tests for `refuceFetchInst` simproc -/
+
+example (h : s.program = GCMGmultV8Program.gcm_gmult_v8_program) :
```
shigoel, line 11: Let's also add a test for a longer program, say SHA512. Is all well or is maxRecDepth reached when the last instruction is fetched?

Tactics/Reflect/FetchAndDecode.lean
```diff
@@ -0,0 +1,58 @@
...
+
+  trace[Sym.reduceFetchInst] "{Lean.checkEmoji} reduced to: {rawInst}"
+
+  -- Now, construct the proof
+  let eqType ← mkEq e (toExpr (some rawInst))
+  trace[Sym.reduceFetchInst] "⚙️ deriving a proof of:\n\t{eqType}"
+
+  let h_program := mkIdent programHyp.userName
+  let eqProofStx ← `(term| by
+    unfold fetch_inst; rw [$h_program:ident]; decide
```
shigoel, line 38: Will `decide` here cause maxRecDepth to be reached for really large programs?

Tests/Tactics/ReduceFetchInst.lean
alexkeizer: I've added tests for the first and last instructions of SHA512. Both are instantaneous (after the simplification described in the next comment)

Tactics/Reflect/FetchAndDecode.lean
alexkeizer: I've simplified this, by extracting out a `fetch_inst_eq_of_prgram_eq_of_map_find` lemma.
The simproc now directly builds an application of that theorem, so that part is instant.

The only risk is that the proof still uses `rfl`, so for truly ginormous programs, kernel reduction might slow down. However, tests for SHA512 are instantaneous ther; it should not be a problem.

alexkeizer:
@shigoel all comments are addressed, another round of review please!

Tactics/Reflect/FetchAndDecode.lean
bollu: @alexkeizer I guess we can have an option that (defaults to) `native_decide`, and uses `decide` if the option (name subject to bikeshed: `minimalTCB := true`) is enabled?

alexkeizer: Post simplification we don't use `decide` at all, only `rfl`. Since it's still instantaneous for SHA512, I don't think we have to worry about this at all

alexkeizer:
Interestingly, things built just fine locally, but it seems CI has some addition checks that ensure we don't import from `Proofs` in `Tests`? I've moved the import of SHA512 to another definition of the program that does live in `Tests` so all should be green now

shigoel:
> Interestingly, things built just fine locally, but it seems CI has some addition checks that ensure we don't import from `Proofs` in `Tests`? I've moved the import of SHA512 to another definition of the program that does live in `Tests` so all should be green now

No such check exists in the CI, but here's something interesting. The file is `Proofs/SHA512/Sha512Program.lean` but the import was `import Proofs.SHA512.SHA512Program` (prior to your latest commit). Apparently, this works locally (for me too), but not in CI (snippet of CI log below).

```
no such file or directory (error code: 2)
  file: ././././Proofs/SHA512/SHA512Program.lean
```

Tactics/Reflect/FetchAndDecode.lean
shigoel: Thanks, @alexkeizer. The addition of `fetch_inst_eq_of_prgram_eq_of_map_find` addresses my concern.

BTW, there are other large programs we can test this on. E.g., see https://github.com/leanprover/LNSym/blob/main/Tests/AES-GCM/AESGCMEncKernelProgram.lean.

alexkeizer:
Aha, Linux is indeed case-sensitive, I did not know MacOS was not. Oh well, this works

shigoel (approved):

PR #83, opened 2024-08-14
feat: spawn tasks for cosimulation to provide (theoretically) $nproc parallelism

bollu:
### Description:

This makes running cosimulation faster. This is nice for faster CI runs, where cosimulation appears to be the bottleneck. 

### Testing:

### License:


bollu:
When taking a walk, I realized that this can be made much cleaner by naming the files on the lean end. the Lean code knows how many tasks have been queued so far, and this can be passed as an argument to both `disasm` as well as `cosim` to name their files.

shigoel:
I like where this is going!

bollu:
Yay, it's in fact faster.

Old:
```
-- https://github.com/leanprover/LNSym/actions/runs/10378815072/job/28735781299#step:8:10474
real 585.45
user 405.94
sys 226.58
```

new (from this commit):

```
-- https://github.com/bollu/LNSym/actions/runs/10394146295/job/28783282692
real 451.05
user 650.40
sys 419.80
```

Arm/Cosim.lean
```diff
@@ -99,7 +99,7 @@ def machine_to_regState (inst : BitVec 32) (str : String) : regState :=
   { inst, gpr, nzcv := flags[0]!, sfp }
 
 /-- Call the `armsimulate` script. -/
-def arm_cosim_test (input : regState) : IO String := do
+def arm_cosim_test (input : regState) (guid : Nat) : IO String := do
```
alexkeizer, line 102: What is `guid` being used for?

Arm/Cosim.lean
```diff
@@ -109,7 +109,8 @@ def arm_cosim_test (input : regState) : IO String := do
   let gprs'   := bitvec_to_hex_list input.gpr
   let flags'  := bitvec_to_hex input.nzcv
   let sfps'   := bitvec_to_hex_list input.sfp
-  let args    := ([inst'] ++ gprs' ++ [flags'] ++ sfps').toArray
+  let instructionStr := ([inst'] ++ gprs' ++ [flags'] ++ sfps')
+  let args    := [toString guid] ++ instructionStr |>.toArray
```
alexkeizer, line 113:
```suggestion
  let args    := toString guid :: instructionStr |>.toArray
```
Seems more readable to me, but tastes differ

Also, what is instructionStr exactly?

Arm/Cosim.lean
```diff
@@ -206,30 +210,52 @@ def regStates_match (input o1 o2 : regState) : IO Bool := do
...
+Use a `guid` to create unique files to ensure that tests do not trample on each other.
+Return `some t` if a test can be produced, and `none` if not.
+- Uses `IO` to try produce a random test that shall be run in a task.
+- Returns `.none` if there is the instruction does not exist on the given architecture.
+- Returns `.some task` upon succeessful creation of the task to randomly test one instruction instance.
+-/
+def mk_one_test_task (verbose : Bool) (logPrefix : String)
+    (fn : IO (Option (BitVec 32))) (guid : Nat) :
+    IO (Option (Task (Except IO.Error Bool))) := do
+  let maybe_inst ← fn
```
alexkeizer, line 232: This seems to create a separate `Task` per test cases. I vaguely remember a discussion that Lean's `Task` can be a bit heavy, and in particular, it's best to not create more `Task`s than CPU cores.

I'm not sure how accurate this memory is w.r.t. the current situation, this might've been quite some time ago.

alexkeizer (commented):
Looks good!

For future work: I had a chat with @bollu: he suggested most time is spent assembling the reference binaries, so it could be an idea to cache those on disk. That would at least speed up local runs of conformance testing (and CI if we jump through the extra hoop of CI caching)

Arm/Cosim.lean
bollu: I elaborated on this in the new commit.

Arm/Cosim.lean
bollu: I rename it to `fileBaseName` so we know that it's the filename that corresponds to the test.

Arm/Cosim.lean
bollu: This is cleaner to implement, and in the glorious future when `Task`s will work as intended (green threads), this should perform even better than it currently does!

bollu:
@alexkeizer comments addressed!

Arm/Cosim.lean
```diff
@@ -109,8 +115,7 @@ def arm_cosim_test (input : regState) (guid : Nat) : IO String := do
   let gprs'   := bitvec_to_hex_list input.gpr
   let flags'  := bitvec_to_hex input.nzcv
   let sfps'   := bitvec_to_hex_list input.sfp
-  let instructionStr := ([inst'] ++ gprs' ++ [flags'] ++ sfps')
-  let args    := [toString guid] ++ instructionStr |>.toArray
+  let args    := [uniqueBaseName] ++ [inst'] ++ gprs' ++ [flags'] ++ sfps' |>.toArray
```
alexkeizer, line 118:
```suggestion
  let args    := [uniqueBaseName, inst'] ++ gprs' ++ [flags'] ++ sfps' |>.toArray
```

bollu: @alexkeizer I disagree, the current ordering is "semantic" since the base name semantically belongs to a different "category"

Arm/Insts/Cosim/armsimulate
```diff
@@ -1,14 +1,22 @@
+#!/usr/bin/env bash
 # Borrowed from the s2n-bignum project
 # https://github.com/awslabs/s2n-bignum/blob/main/arm/proofs/simulator.c
 # Author(s): John Harrison, Juneyoung Lee
+# Usage: armsimluate <output-file-name> <instruction in GNU assembly syntax>
```
shigoel, line 5:
```suggestion
# Usage: armsimulate <output-file-name> <32-bit instruction>
```

Arm/Cosim.lean
```diff
@@ -98,8 +98,14 @@ def machine_to_regState (inst : BitVec 32) (str : String) : regState :=
   let sfp := List.map (fun x => (BitVec.ofNat 64 x.toNat!)) (strs.take 64)
   { inst, gpr, nzcv := flags[0]!, sfp }
 
-/-- Call the `armsimulate` script. -/
-def arm_cosim_test (input : regState) : IO String := do
+/--
+Call the `armsimulate <uniqueBaseName> <register>` script to
```
shigoel, line 102:
```suggestion
Call the `armsimulate <uniqueBaseName> <input>` script to
```

Arm/Cosim.lean
```diff
@@ -98,8 +98,14 @@ def machine_to_regState (inst : BitVec 32) (str : String) : regState :=
...
+build an executable and report the results of executing <register>.
```
shigoel, line 103:
```suggestion
build an executable and report the results of executing <input>.
```

Arm/Cosim.lean
```diff
@@ -206,30 +218,53 @@ def regStates_match (input o1 o2 : regState) : IO Bool := do
...
   let model_st := model_to_regState inst model
   regStates_match input machine_st model_st
 
+/--
+Make a task for running a single test.
+Use a `uniqueBaseName` to create unique files to ensure that tests do not
+clobber each other's state.
+Return `some t` if a test can be produced, and `none` if not.
+- Uses `IO` to try produce a random test that shall be run in a task.
+- Returns `.none` if there is the instruction does not exist on the given architecture.
```
shigoel, line 235:
```suggestion
- Returns `.none` if the instruction does not exist on the given architecture.
```

Arm/Cosim.lean
```diff
@@ -206,30 +218,53 @@ def regStates_match (input o1 o2 : regState) : IO Bool := do
...
+- Returns `.some task` upon succeessful creation of the task to randomly test one instruction instance.
```
shigoel, line 236:
```suggestion
- Returns `.some task` upon successful creation of the task to randomly test one instruction instance.
```

Arm/Insts/Cosim/common.sh
shigoel, line 1: Where's this script used?

Arm/Cosim.lean
```diff
@@ -206,30 +218,53 @@ def regStates_match (input o1 o2 : regState) : IO Bool := do
...
+-/
+def mk_one_test_task (verbose : Bool) (logPrefix : String)
+    (fn : IO (Option (BitVec 32))) (uniqueBaseName : String) :
+    IO (Option (Task (Except IO.Error Bool))) := do
+  let maybe_inst ← fn
+  match maybe_inst with
+  | .none => return .none
+  | .some inst => IO.asTask do
+    let ret ← one_test inst uniqueBaseName
+    -- NOTE: this is broken, since it assumes that we know the file name that we are disassembling x(
```
shigoel, line 246: Old, irrelevant comment?

Arm/Insts/Cosim/armsimulate
shigoel: The script doesn't really expect an instruction in assembly syntax, right? It expects a 32-bit value?

shigoel: Also, since you're writing the usage here, might as well talk about all the arguments :)  (copied from `arm_cosim_test` in Arm/Cosim.lean):

```
-- Input args for the armsimulate script:
  --  first, the 32-bit instruction
  --  then 31 64-bit GPRs (no SP)
  --  then 1 4-bit Flags
  --  then 32*2 64-bit SPFs
```

Arm/Insts/Cosim/disasm.sh
```diff
@@ -18,7 +20,7 @@ if [ -e instance.o ]; then
 
     # Do not print the address -- just the instruction bytes and its
     # corresponding disassembly.
-    snip_inst="awk '{\$1=\"\"; print}'"
+    snip_inst="awk '{\$2=\"\"; print}'"
```
shigoel, line 23:
```suggestion
    snip_inst="awk '{\$1=\"\"; print}'"
```
Why `$2`? We'd like to see the 32-bit instruction value in the output, not the offset.

shigoel:
> Looks good!
> 
> For future work: I had a chat with @bollu and he suggested most time is spent assembling the reference binaries, so it could be an idea to cache those on disk. That would at least speed up local runs of conformance testing (and CI if we jump through the extra hoop of CI caching)

Another idea, which perhaps is out of scope right now, would be to do away with all this file I/O. We could define one C/Asm function per instruction (which could be autogenerated from a list of instructions), and use shared object libraries to directly call these external functions in Lean.

Arm/Insts/Cosim/disasm.sh
bollu: @shigoel it needs to be `$2` because we now take an additional argument which is the file name to read from (`INFILE`), no?

Arm/Insts/Cosim/armsimulate
bollu: Fixed, thanks!

Arm/Insts/Cosim/disasm.sh
bollu: Fixed, d'oh, it's an `awk` $1 not a shell `$1`

Arm/Cosim.lean
bollu: Thanks, cleaned up

Arm/Insts/Cosim/common.sh
bollu: Fixed, killed it as it's un-necessary

shigoel (approved):
This is nice!

PR #87, opened 2024-08-14
fix: use simprocs in step-theorem generation

alexkeizer:
### Description:

Some stepi-theorems were being generated which still included elements like `decode_bit_masks 0#1 3#6 0#6 false`.
After investigation, I noticed simprocs were *not* being passed to `simp` in `StepThms`, so I fixed that.

### Testing:

`make all` ran locally

### License:

### DIFF
--- a/Proofs/Experiments/AbsVCG.lean
+++ b/Proofs/Experiments/AbsVCG.lean
@@ -200,8 +200,6 @@ theorem program_effects_lemma (h_pre : abs_pre s0)
     simp only [run, h_step_2]
   replace h_step_2 : s2 = stepi s1 := h_step_2.symm
   rw [program.stepi_0x4005d4 s1 s2 h_s1_program h_s1_pc h_s1_err] at h_step_2
-  -- (FIXME) abs_stepi_0x4005d4 should have reduced `decode_bit_masks`.
-  simp only [reduceDecodeBitMasks] at h_step_2
   have h_s2_program : s2.program = program := by
     simp only [h_step_2, h_s1_program,
                state_simp_rules, minimal_theory, bitvec_rules]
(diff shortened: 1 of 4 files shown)


Proofs/Experiments/AbsVCG.lean
```diff
@@ -274,6 +272,7 @@ theorem program_effects_lemma (h_pre : abs_pre s0)
             BitVec.truncate 64 1#32)) := by
     simp (config := {decide := true}) only [h_step_4, h_step_3, h_step_2, h_step_1,
                                             state_simp_rules]
+    rfl
```
alexkeizer, line 275: Not sure why this rfl is now needed, but the proof broke without it

bollu:
LGTM

Proofs/SHA512/Sha512DecodeExecLemmas.lean
```diff
@@ -25,8 +25,7 @@ info: sha512_program.exec_0x126c90 (s : ArmState) :
       (ArmInst.BR (BranchInst.Compare_branch { sf := 1#1, _fixed := 26#5, op := 1#1, imm19 := 523804#19, Rt := 2#5 }))
       s =
     w StateField.PC
-      (if decide (r (StateField.GPR 2#5) s = BitVec.zero (if 1#1 = 1#1 then 64 else 32)) = false then
-        r StateField.PC s + BitVec.signExtend 64 (523804#19 <<< 2)
+      (if decide (r (StateField.GPR 2#5) s = BitVec.zero 64) = false then r StateField.PC s + 18446744073709549680#64
```
shigoel, line 28: Nice! I've been so tired of seeing ground terms like `(if 1#1 = 1#1 then 64 else 32)`. Guess the wrong way of passing in `simprocs` was the culprit all along.

Proofs/Experiments/AbsVCG.lean
shigoel: It's brittle anyway -- I'll redo all of it later. Thanks!

shigoel (approved):
LGTM, thanks, @alexkeizer!

PR #92, opened 2024-08-16
Rework step theorem generation to be faster and cache intermediate results to the environment

alexkeizer:
### Description:

Forewarning: the diff of this one is rather large.

I've implemented a new version of step theorem generation, culminating in the `genStepEqTheorem` function.

In particular:
- We get rid of the different fetch/decode/exec intermediate lemma, instead opting to generate the final step theorem in one go. The bottleneck seems to be kernel checking, so by reducing the number of theorems sent to the kernel we achieve a good speedup (SHA512 used to take about 55 seconds for the three generation commands combined, now it's about 35 seconds).
- In the process, we build a `ProgramInfo` struct, which holds a bunch of interesting expressions that further proof automation could exploit. This programInfo is stored in a persistent environment extension, so that it is persisted in the olean files (hence, making it available to downstream files).
- While building the previous, I've moved some definitions around and did other refactors: happy to split those off into their own PR if that makes reviewing easier.

EDIT: split off #93 for the `#time` command

### Testing:

`make all` succeeded

### License:


Arm/Exec.lean
```diff
@@ -166,3 +166,13 @@ theorem run_onestep (s s': ArmState) (n : Nat) (h_nonneg : 0 < n):
...
+
+/-- helper lemma for automation -/
+theorem stepi_eq_of_fetch_inst_of_decode_raw_inst
+    (s : ArmState) (addr : BitVec 64) (rawInst : BitVec 32) (inst : ArmInst)
+    (h_err    : read_err s = .None)
+    (h_pc     : read_pc s = addr)
+    (h_fetch  : fetch_inst addr s = some rawInst)
+    (h_decode : decode_raw_inst rawInst = some inst) :
+    stepi s = exec_inst inst s := by
+  simp [stepi, *]
```
bollu, line 178: simp only, please, even terminal 😞

Tactics/Common.lean
```diff
@@ -88,13 +91,13 @@ that additionally recognizes:
...
-      if h : n = 2^n' then
-        return ⟨n', .ofFin (Fin.cast h i)⟩
-      else
-        failure
+    | BitVec.ofFin w i => OptionT.run do
+      let w ← getNatValue? w
+      let v ← do
+        match_expr i with
+          | Fin.mk _n v _h  => getNatValue? v
+          | _               => pure (←getFinValue? i).2.val
```
bollu, line 99:
```suggestion
          | _               => pure (← getFinValue? i).2.val
```

Tactics/Common.lean
```diff
@@ -115,21 +118,54 @@ which was obtained by reducing:\n\t{e}"
...
-  if e.hasFVar then
+  if e.hasFVar || e.hasMVar then
     throwError "Expected a ground term, but {e} has free variables"
 
-  if let some ⟨n, x⟩ ← _root_.getBitVecValue? e then
-    if h : n = w then
-      return x.cast h
-    else
-      throwError "Expected a bitvector of width {w}, but\n\t{e}\nhas width {n}"
+  let some ⟨n, x⟩ ←_root_.getBitVecValue? e
```
bollu, line 124:
```suggestion
  let some ⟨n, x⟩ ← _root_.getBitVecValue? e
```

Tactics/Common.lean
```diff
@@ -115,21 +118,54 @@ which was obtained by reducing:\n\t{e}"
...
-  if (←isDefEq e e') then
-    return BitVec.ofNat w (← reflectNatLiteral x)
+  if h : n = w then
+    return x.cast h
   else
-    throwError "Failed to unify, expected:\n\t{e'}\nbut found:\n\t{e'}"
+    throwError "Expected a bitvector of width {w}, but\n\t{e}\nhas width {n}"
+
+
+/-! ## Hypothesis types -/
```
bollu, line 133:
```suggestion
    throwError "Expected a bitvector of width {w}, but\n\t{e}\nhas width {n}"

/-! ## Hypothesis types -/
```

stray newline

Tactics/Common.lean
```diff
@@ -115,21 +118,54 @@ which was obtained by reducing:\n\t{e}"
...
+namespace SymContext
+
+/-- `h_err_type state` returns an Expr for `r .ERR <state> = .None`,
+the expected type of `h_err` -/
+def h_err_type (state : Expr) : Expr :=
+  mkAppN (mkConst ``Eq [1]) #[
+    mkConst ``StateError,
+    mkApp2 (.const ``r []) (.const ``StateField.ERR []) state,
+    .const ``StateError.None []
+  ]
```
bollu, line 143:
```suggestion
/-- `h_err_type state` returns an Expr for `r .ERR <state> = .None`,
the expected type of `h_err` -/
def h_err_type (state : Expr) : Expr :=
  mkAppN (mkConst ``Eq [1]) #[
    mkConst ``StateError,
    mkApp2 (.const ``r []) (.const ``StateField.ERR []) state,
    .const ``StateError.None []
  ]
```

If I may be annoying, can we name this according to the term we build? So something like `h_State_error_eq_none`? In particular, this seems hard to audit when the definition changes.

I admit, a long name doesn't make me happy 😢

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,318 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
+  /-- the raw instruction, as a bitvector -/
+  rawInst : BitVec 32
+
+  /-- the decoded instruction, as a normalized(!) `Expr` of type `ArmInst`.
+  That is, `decode_raw_inst <rawInst>` should be def-eq to `some <decodeInst>`.
+
+  NOTE: a `none` value indicates that this expression has not been computed yet,
+  *not* that `decode_raw_inst` returned `none`-- we assume that
+  all instructions for which we generate an `InstInfo` are well-formed -/
+  decodedInst? : Option Expr
```
bollu, line 33: Consider making a different inductive for this?

```lean
inductive OnDemand (a : Type) where 
| NotYetDemanded
| Computed (v : a)
```

This way, the semantics are utterly clear.

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,318 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
+    ```lean
+    ∀ s (h_program : s.program = <program>) (h_pc : read_pc s = <PC>)
+        (h_err : read_err s = .None),
+      exec_inst s = <sem> s
+    ```
+  - `proof` is a proof of type `type`
+
+  Otherwise, if the field is `none`, this indicates that the relevant
+  expressions have not been computed/cached yet. -/
+  instSemantics? : Option (Expr × Expr × Expr)
```
bollu, line 48:
```suggestion
  instSemantics? : OnDemand (Expr × Expr × Expr)
```

Same suggestion as above, switch to `OnDemand`.

Tactics/Reflect/ProgramInfo.lean
bollu, line 33:
```suggestion
  decodedInst? : OnDemand Expr
```

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,318 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
         let_expr Prod.mk _ _ addr inst := hd'
           | throwError "expected `{hd}` to reduce to an application of `Prod.mk`, found:\n\t{hd'}"
 
-        let addr ← reflectBitVecLiteral 64 addr
-        let inst ← reflectBitVecLiteral 32 inst
+        let addr ← reflectBitVecLiteral 64 (← instantiateMVars addr)
+        let inst ← reflectBitVecLiteral 32 (← instantiateMVars inst)
 
-        let rawProgram := rawProgram.insert addr inst
+        let rawProgram := instructions.insert addr ⟨inst, none, none⟩
```
bollu, line 142:
```suggestion
        let rawProgram := instructions.insert addr (InstInfo.ofRawInst inst)
```

and make a function that builds an `InstInfo`?

Tactics/Common.lean
alexkeizer: Note that the various uses of `h_err_type` are pretty strongly tied together; if the expression we want at one of it's callsite changes we almost surely want the same change at all the other callsites. Does that alleviate your concerns re:auditability

bollu (commented):
I gave it a read, but I must admit I am fading a little at 3PM on a Friday, so I will come back to this on Monday. Overall, the patch looks LGTM, but I may have nits in `StepThms.lean`.

Tactics/Reflect/ProgramInfo.lean
alexkeizer: Added

Tactics/Reflect/ProgramInfo.lean
alexkeizer: I've made the other field optional, so this now reads `{ rawInst }`, does that spark joy?

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,330 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
+  rawInst : BitVec 32
+
+  /-- the decoded instruction, as a normalized(!) `Expr` of type `ArmInst`.
+  That is, `decode_raw_inst <rawInst>` should be def-eq to `some <decodeInst>`.
+  -/
+  decodedInst? : OnDemand Expr :=
+    .notYetComputed
+
+  /-- if `instSemantics?` is `⟨sem, type, proof⟩`, then
+  - `sem` is the instruction semantics, as a normalized(!) expression of type
```
shigoel, line 44: Help me out here: what does `normalized` really mean?

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,330 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
+    ```
+  - `proof` is a proof of type `type`
+  -/
+  instSemantics? : OnDemand (Expr × Expr × Expr) :=
+    .notYetComputed
+
 structure ProgramInfo where
-  rawProgram : HashMap (BitVec 64) (BitVec 32)
+  name : Name
+  instructions : HashMap (BitVec 64) InstInfo
```
shigoel, line 59: This encapsulation is gorgeous, Alex!

Tactics/Reflect/ProgramInfo.lean
alexkeizer: I guess simplified would be a better phrasing. I meant to emphasize that this should be a sequence of `w`/`write_mem`s (except it's not always, because sometimes we have annoying `if`s in there still)

shigoel: Gotcha. Yeah, maybe add this detail in the doc.

BTW, not all `if`s can be avoided. E.g., what about a conditional branch instruction? Should we generate two sets of semantics there, one where the condition holds and one where it doesn't?

Arm/Exec.lean
```diff
@@ -166,3 +166,13 @@ theorem run_onestep (s s': ArmState) (n : Nat) (h_nonneg : 0 < n):
...
+
+/-- helper lemma for automation -/
+theorem stepi_eq_of_fetch_inst_of_decode_raw_inst
+    (s : ArmState) (addr : BitVec 64) (rawInst : BitVec 32) (inst : ArmInst)
+    (h_err    : read_err s = .None)
+    (h_pc     : read_pc s = addr)
+    (h_fetch  : fetch_inst addr s = some rawInst)
+    (h_decode : decode_raw_inst rawInst = some inst) :
+    stepi s = exec_inst inst s := by
+  simp only [stepi, h_err, h_pc, h_fetch, h_decode]
```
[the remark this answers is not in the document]
alexkeizer: So the `read_err`/`read_pc` won't show up in the goal state because of the way this lemma is used, but you're right that it's cleaner to just use `r` if it doesn't matter, so I've changed it

Tactics/Common.lean
```diff
@@ -46,6 +46,9 @@ def filterDeclsWithPrefix (lctx : LocalContext) (pfxUserName : Name)
           | some decl => String.isPrefixOf (toString pfxUserName)
                                            (toString decl.userName))))
 
+def BitVec.toHexWithoutLeadingZeroes {w} (x : BitVec w) : String :=
+  (Nat.toDigits 16 x.toNat).asString
+
```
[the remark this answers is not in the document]
alexkeizer: I mirrored what `toHex` does (which also does not have the leading `0x`), that felt safest.
You are right that everywhere we use it we append `0x`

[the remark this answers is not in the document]
shigoel: I kinda don't like that about `toHex`. I've found that without the base indicators, it's easy to misinterpret what `20` means. Is it 20 or 32? *shrug*

Tactics/Reflect/ProgramInfo.lean
alexkeizer: It now reads:
```
  - `sem` is the instruction semantics, as a simplified expression of type
      `ArmState → ArmState`.

    That is, we've ran `simp` on `sem` with our dedicated simp-sets in the hopes
    of obtaining only a sequence of `w` and `write_mem`s to the initial state.
    However, note that some instructions might have conditional behaviour, in
    which case `sem` might still contain `if`s
```

> BTW, not all ifs can be avoided. E.g., what about a conditional branch instruction? Should we generate two sets of semantics there, one where the condition holds and one where it doesn't?

Good question: I'm not sure what the exact shape should be, but preprocessing to get different semantics expressions for different situations seems like a good idea (to be implemented at some point in the future).

Tactics/Reflect/ProgramInfo.lean
```diff
@@ -17,81 +17,330 @@ Furthermore, we define a persistent env extension to store `ProgramInfo` in.
...
+
+--------------------------------------------------------------------------------
+
+namespace ProgramInfo
+
+/-- The expression `mkConst pi.name`,
+i.e., an expression of this program referred to by name -/
+def expr (pi : ProgramInfo) : Expr := mkConst pi.name
+
+def getInstInfoAt? (pi : ProgramInfo) (addr : BitVec 64) :
```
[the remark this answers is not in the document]
alexkeizer: Indeed, but I feel like we'd have to do that anyway. If we had to run an `omega`-like automation every step to determine what the instruction at the current PC actually is, I'd expect that to tank performance

Tactics/Common.lean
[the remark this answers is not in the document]
alexkeizer: Fair enough, I'm happy to change it

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+    mkApp2 (.const ``Eq.refl [1])
+      (mkApp (.const ``Option [0]) <|
+        mkApp (.const ``BitVec []) (toExpr 32))
+      someRawInst
+  ]
+
+-- /-! ## `reduceDecodeInst` -/
+
+/-- `canonicalizeBitVec e` recursively walks over expression `e` to convert any
+occerences of:
```
shigoel, line 68:
```suggestion
occurrences of:
```

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+  `BitVec.ofFin w (Fin.mk x _)`
```
shigoel, line 69: I feel like I'm missing something. Where could these `BitVec.ofFin ...` terms come from? From `simp/ground` (which I've personally run into)? Anywhere else?

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+    | _ => fallback
+  where
+    fallback : MetaM Expr := do
+      let fn   := e.getAppFn
+      let args  ← e.getAppArgs.mapM canonicalizeBitVec
+      return mkAppN fn args
+
+/-- Given an expr `rawInst` of type `BitVec 32`,
+return an expr of type `Option ArmInst` representing what `rawInst` decodes to.
+The resulting expr is guaranteed to be def-eq to `fetch_inst $rawInst` -/
```
shigoel, line 95:
```suggestion
The resulting expr is guaranteed to be def-eq to `decode_raw_inst $rawInst` -/
```

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+abbrev SymM := ProgramInfoT <| MonadCacheT SymM.CacheKey Expr MetaM
+
+@[inherit_doc ProgramInfoT.run]
+abbrev SymM.run (name : Name) (k : SymM α) (persist : Bool := true) : MetaM α :=
+  MonadCacheT.run <| ProgramInfoT.run name k persist
+
+open SymM in
+/-- Given a (reflected) raw instruction,
+return an expr of type `Option ArmInst` representing what `rawInst` decodes to.
+The resulting expr is guaranteed to be def-eq to `fetch_inst $rawInst`.
```
shigoel, line 115:
```suggestion
The resulting expr is guaranteed to be def-eq to `decode_raw_inst $rawInst`.
```

Tactics/Common.lean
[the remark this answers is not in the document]
alexkeizer: Hmm, we I changed `getBitVecString?` to use `toHexWithoutLeadingZeroes`, but, the former is used in `getStateFieldString?`, where we don't really want the `0x` prefix. 

```
def getStateFieldString? (e : Expr) : MetaM (Option String) := OptionT.run do
  match_expr e with
  | StateField.GPR iExpr  => return ("x" ++ (← getBitVecString? iExpr))
  | StateField.SFP iExpr  => return ("q" ++ (← getBitVecString? iExpr))
  | StateField.PC         => return "pc"
  | StateField.FLAG pExpr => getPFlagString? pExpr
  | StateField.ERR        => return "err"
  | _                     => panic! s!"[getStateFieldName?] Unexpected expression: {e}"
```

So, I'll stick with the old behaviour. I have moved it to `Arm/BitVec.lean`!

Tactics/StepThms.lean
shigoel: `reduce`?

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+
+      let (ctx, simprocs) ← do
+        let localDecls ← do
+          let hs := #[h_pc, h_err]
+          pure <| hs.filterMap (← getLCtx).findFVar?
+        LNSymSimpContext
+          (config := {decide := true, ground := false})
+          (simp_attrs := #[`minimal_theory, `bitvec_rules, `state_simp_rules])
+          (decls := localDecls)
+          (decls_to_unfold := #[``exec_inst, ``read_pc, ``read_err])
```
shigoel, line 175:
```suggestion
          (decls_to_unfold := #[``exec_inst])
```
`read_pc/err` are in `state_simp_rules`

Tactics/StepThms.lean
```diff
@@ -33,11 +41,190 @@ initialize registerTraceClass `gen_step.print_names
...
+
+open ProgramInfoT InstInfoT
+
+/-! ## reduceStepiToExecInst -/
+
+/-- Given a program and an address, and optionally the corresponding
+raw and decoded instructions, construct and return first the expression:
+```
+∀ {s} (h_program : s.program = <progam>) (h_pc : read_pc s = <addr>)
+  (h_err : read_err s = .None),
```
shigoel, line 130:
```suggestion
∀ {s} (h_program : s.program = <progam>) (h_pc : r .PC s = <addr>)
  (h_err : r .ERR s = .None),
```

shigoel: Something to keep in mind: some LDST instructions expect the SP to be aligned (e.g., https://github.com/leanprover/LNSym/blob/6d1ef32d9d253b8b7667edfaa08f38ea82da2dc1/Arm/Insts/LDST/Reg_unscaled_imm.lean#L28). We may need that assumption for such instructions.

Tactics/StepThms.lean
alexkeizer: It comes from `reduce` indeed, the idea being that instead of painstakingly crafting a simpset that unfold everything but `BitVec.ofNat`, we rather reduce and then refold the `BitVec`s into `BitVec.ofNat`

Tactics/StepThms.lean
alexkeizer: Yes, this is the same problem as you mentioned for branch: the exact conditions needed for a semantics differ. For now, such conditions will remain present as `if`s in the simplified semantics, which are then reduced away by `sym_n`.

In the future, I'd like to make this smarter, by introducing a more elaborate datastructure that tracks exactly which assumptions are needed, and fully reduce/reflect for that set of assumptions. I figured this PR is already big enough as it is, so I didn't want to go too deep into that train of thought yet

Tactics/StepThms.lean
shigoel: Sounds good --- maybe add this in as a comment.

shigoel:
@alexkeizer I'll give this PR another read on Mon, with a fresher brain and hopefully a broader outlook. This is so exciting!

Tactics/StepThms.lean
alexkeizer: Done, it now reads:
```
/-- `canonicalizeBitVec e` recursively walks over expression `e` to convert any
occurrences of:
  `BitVec.ofFin w (Fin.mk x _)`
to the canonical form:
  `BitVec.ofNat w x` (i.e., `x#w`)

Such expressions tend to result from using `reduce` or
`simp` with `{ground := true}`.
You can call `canonicalizeBitVec` after these functions to ensure you don't
needlessly expose `BitVec` internal details -/
```

alexkeizer:
For illustration, on the [latest commit in main](https://github.com/leanprover/LNSym/actions/runs/10425720489/job/28879931756), the linux CI built the Proofs module in:
```
✔ [449/450] Built Proofs
Build completed successfully.
real 134.36
user 360.74
sys 16.15
```

Whereas, on [the latest commit of this PR](https://github.com/leanprover/LNSym/actions/runs/10427155835/job/28881359157?pr=92), it only took: 
```
✔ [448/449] Built Proofs
Build completed successfully.
real 104.34
user 263.48
sys 15.58
```

Tactics/Common.lean
shigoel: @alexkeizer @bollu What's the conclusion of this discussion? I'm happy with the way things stand right now. If you both are happy, hit "resolve" and we're ready to merge this PR!

shigoel (approved):

Tactics/Common.lean
alexkeizer: I'm happy with the current naming; I'd be a bit hesitant to adopt longer names like @bollu suggests, so I'd like to merge it as is, if Sid agrees

alexkeizer: I'm resolving this after a chat with @bollu

alexkeizer:
@shigoel this is now ready to merge, but I unfortunately cannot hit that button myself

PR #93, opened 2024-08-16
feat: copy `#time` command from Mathlib

alexkeizer:
### Description:

Copy the `#time` command, with some adaptations, from Mathlib, since it's very useful in getting rough performance numbers for automation.

License-wise, this should be fine, since both are Apache licensed, but we're waiting on an official answer from Leo on that.

EDIT: Leo confirmed it's fine, but also suggested asking Kim if it's worth upstreaming to core, so [I've pinged them](https://leanprover.zulipchat.com/#narrow/stream/424609-lean-at-aws/topic/Moving.20.60.23time.60.20to.20core). Let's wait for the result of that

### License:

### DIFF
--- a/Tactics/Time.lean
+++ b/Tactics/Time.lean
@@ -1,0 +1,33 @@
+/-
+Copyright (c) 2021 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+Taken from: https://github.com/leanprover-community/mathlib4/blob/318082b0bccc3abd9d654496f7b60267f277d5fd/Mathlib/Util/Time.lean#L28-L33
+-/
+import Lean
+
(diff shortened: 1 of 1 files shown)


bollu:
LGTM, yoink it in. Maybe ask Kim if we can toss it into core :)

shigoel:
When we bump the toolchain to at least [nightly-2024-08-15](https://github.com/leanprover/lean4-nightly/releases/tag/nightly-2024-08-15) where `#time` was upstreamed, we can snuff out this PR.

alexkeizer:
We can't really update the toolchain until LeanSAT gets merged completely (or the library gets updated, but that's unlikely to happen). Luckily, the process seems to be going quite quickly: [PR 4 out of 6](https://github.com/leanprover/lean4/pull/5074) just got merged

PR #94, opened 2024-08-16
 chore: remove old versions of the sym tactic, and rename the new `sym1_n` tactic to `sym_n`

alexkeizer:
Depends on #92

### Description:

Relatively minor cleanup that removes the old versions of symbolic simulation in favor of the new `sym1_n` tactic, which I subsequently renamed to the more intuitive `sym_n`, now that this name is available.

### Testing:

`make all` ran locally

### License:


Tactics/Sym.lean
```diff
@@ -211,6 +141,18 @@ elab "sym1_n" n:num s:(sym_at)? : tactic =>
...
+    -- Check that step theorems have been pre-generated
+    try
+      let pc := c.pc.toHexWithoutLeadingZeroes
+      let step_thm := Name.str c.program ("stepi_eq_0x" ++ pc)
+      let _ ← getConstInfo step_thm
+    catch err =>
+      throwErrorAt err.getRef "{err.toMessageData}\n
+Did you remember to generate step theorems with:
+  #generateStepEqTheorems {c.program}"
+-- TODO: can we make this error ^^ into a `Try this:` suggestion that
```
shigoel, line 153: Ooo yes please!

Tests/Tactics/Sym.lean
```diff
@@ -7,15 +7,15 @@ import Proofs.«AES-GCM».GCMGmultV8Sym
...
 #guard_msgs in example (s0 sf : ArmState)
     (h_s0_program : s0.program = gcm_gmult_v8_program)
     (h_s0_pc : read_pc s0 = gcm_gmult_v8_program.min)
     (h_run : sf = run gcm_gmult_v8_program.length s0) :
     read_err sf = .None := by
   simp (config := {ground := true}) only [Option.some.injEq] at h_s0_pc h_run
-  sym1_n 1
+  sym_n 1
   · sorry
   · exact (sorry : CheckSPAlignment s0)
```
shigoel, line 20: For a subsequent PR: if `CheckSPAlignment s0` is not an original hyp., then don't generate this proof obligation.

Tests/Tactics/Sym.lean
```diff
@@ -7,15 +7,15 @@ import Proofs.«AES-GCM».GCMGmultV8Sym
...
   · exact (sorry : read_err s0 = .None)
```
shigoel, line 21: Wondering why the behavior is different for a missing `read_pc` hyp. vs `read_err` hyp. IIRC, in the former, we get an error and in the latter, we get a new subgoal.

Tactics/SymContext.lean
```diff
@@ -240,32 +224,82 @@ where
...
+    replaceMainGoal (goal :: newGoals)
+    return ctx
+
+/-- change the type (in the local context of the main goal)
+of the hypotheses tracked by the given `SymContext` to be *exactly* of the shape
+described in the relevant docstrings.
+
+That is, (un)fold types which were definitionally, but not syntactically,
+equal to the expected shape. -/
+def changeHypothesisTypes (c : SymContext) : TacticM Unit := withMainContext do
```
shigoel, line 280: Perhaps name this to be more indicative of its behavior? Maybe `normalize` or `canonicalize` instead of change?

Tests/Tactics/Sym.lean
alexkeizer: As discussed: we need the `read_pc` to be present, since we infer the value of the PC, hence, which stepi lemma to apply, from its type. 
`read_err`, OTOH, does not give us any information, we only need to know that it's true. Thus, if it's not present, we can create a new goal for it.

If we wanted to create a new goal for `read_pc`, we'd have to somehow guess the value of the PC. We *could* do that, e.g., by taking `Program.min` as a logical starting point, but this is necessarily heuristics-based, so I figured throwing an error is simpler (same thing for `h_program`, except that it'd be even harder to make any sensible guess for the right program)

Tests/Tactics/Sym.lean
alexkeizer: It's on my radar! (although a tiny bit more complicated than I thought: I removed the addition of the goals from `SymContext`, but apparently `intro_fetch_decode_lemmas` *also* introduces an `SPAlignment` goal if it can't find the hyp, I'll have to dig into that tactic)

Tactics/SymContext.lean
alexkeizer: I went with canonicalize!

alexkeizer:
@shigoel, this should be ready to merge now

Tactics/Sym.lean
```diff
@@ -12,6 +12,9 @@ import Tactics.SymContext
 
 import Lean
 
+example : True := by
+  simp?
```
shigoel, line 16: @alexkeizer ??

alexkeizer: Oops

alexkeizer: @shigoel fixed, sorry about that

shigoel (approved):
Thanks!

PR #95, opened 2024-08-19
Respect Apple's ABI for cosimulations

shigoel:
### Description:

Workaround to avoid clobbering `x18` and `x29` on Apple's Arm machines.

### Testing:

`make all` ran on my M3 and a Graviton2 machine.

### License:

### DIFF
--- a/Arm/Insts/Common.lean
+++ b/Arm/Insts/Common.lean
@@ -13,6 +13,34 @@ section Common
 open BitVec
 
 ----------------------------------------------------------------------
+/-- `GPRIndex.rand` picks a safe GPR index (i.e., one not reserved on
+Apple platforms). Use this function instead of `BitVec.rand` to pick
+an appropriate random index for a destination GPR during
+cosimulations. We say "destination" because using reserved registers
+as source operands does not violate the Apple ABI.
+
+For details, see
+https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers
+-/
+partial def GPRIndex.rand (lo := 0) (hi := 31) : 
+  IO (BitVec 5) := do
+  let feat_check ←
+  IO.Process.output
+      { cmd  := "Arm/Insts/Cosim/platform_check.sh",
+        args := #["-d"] }
+  if feat_check.exitCode == 0 then
+    BitVec.rand 5 lo hi
+  else 
+    go lo hi
+  where go (lo hi : Nat) : IO (BitVec 5) := do
+    let ans ← BitVec.rand 5 lo hi
+    -- GPRs 18 and 29 are reserved on Apple Arm platforms.
+    if ans ∈ [18#5, 29#5] then
+      go lo hi
+    else
+      pure ans
+
+----------------------------------------------------------------------
 
 def AddWithCarry (x : BitVec n) (y : BitVec n) (carry_in : BitVec 1) :
   (BitVec n × PState) :=
@@ -590,5 +618,7 @@ inductive MemOp where
   | MemOp_STORE : MemOp
   | MemOp_PREFETCH : MemOp
 deriving DecidableEq, Repr
+
+----------------------------------------------------------------------
 
 end Common
--- a/Arm/Insts/Common.lean
+++ b/Arm/Insts/Common.lean
@@ -21,6 +21,10 @@ as source operands does not violate the Apple ABI.
 
 For details, see
 https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers
+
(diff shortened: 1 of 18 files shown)


Arm/Insts/Common.lean
```diff
@@ -12,6 +12,35 @@ section Common
...
+https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers
+-/
+partial def GPRIndex.rand (lo := 0) (hi := 31) :
+  IO (BitVec 5) := do
+  let darwin_check ←
+  IO.Process.output
+      { cmd  := "Arm/Insts/Cosim/platform_check.sh",
+        args := #["-d"] }
+  if darwin_check.exitCode == 1 then
+    go lo hi
```
bollu, line 32:
```suggestion
  let darwin_check ←
  IO.Process.output
      { cmd  := "Arm/Insts/Cosim/platform_check.sh",
        args := #["-d"] }
  if darwin_check.exitCode == 1 then
    go lo hi
```

This feels expensive to recompute on each call; can we cache this on the caller side instead?

Arm/Insts/DPSFP/Conversion_between_FP_and_Int.lean
```diff
@@ -103,7 +103,14 @@ partial def Conversion_between_FP_and_Int_cls.fmov_general.rand : IO (Option (Bi
         rmode  := rmode,
         opcode := opcode,
         Rn := ← BitVec.rand 5,
-        Rd := ← BitVec.rand 5 }
+        Rd := ← (if (lsb opcode 0) = 1#1 then
+                -- FPConvOp.FPConvOp_MOV_ItoF
+                -- GPR used as a source operand.
```
pennyannn, line 108: I recommend changing the comment into `Destination operand is a SIMD&FP register`.

Arm/Insts/DPSFP/Conversion_between_FP_and_Int.lean
```diff
@@ -103,7 +103,14 @@ partial def Conversion_between_FP_and_Int_cls.fmov_general.rand : IO (Option (Bi
...
+                  BitVec.rand 5
+                else
+                -- FPConvOp.FPConvOp_MOV_FtoI
+                -- GPR used as a destination operand.
```
pennyannn, line 112: Likewise, `Destination operand is a GPR register`.

bollu:
LGTM, I'm a little hesitant about the cost of spinning up a new process and checking the platform each time we want a GPR. I would prefer if we refactor this into a `abbrev SampleM := ReaderT Config IO`, where 

```

namespace Sample
structure Config where
    isDarwin : Bool

def SampleM := ReaderT Config IO

def Config.setup : IO Config := do
    pure { isDarwin := true } /- do the process querying here-/

def SampleM.run (m : SampleM a) : IO a := do
  ReaderT.run m (← Config.setup)

end Sample
```

bollu:
I'm happy to perform the refactor I propose as a follow up PR if you like, @shigoel :)

Arm/Insts/Common.lean
```diff
@@ -12,6 +12,34 @@ section Common
 
 open BitVec
 
+----------------------------------------------------------------------
+/-- `GPRIndex.rand` picks a safe GPR index (i.e., one not reserved on
+Apple platforms). Use this function instead of `BitVec.rand` to pick
+an appropriate random index for a destination GPR during
+cosimulations. We say "destination" because using reserved registers
```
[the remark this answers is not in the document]
shigoel: Reading reserved machine components is usually not a problem; modifying them is. I can't find a reference at the moment though, sorry.

shigoel:
> LGTM, I'm a little hesitant about the cost of spinning up a new process and checking the platform each time we want a GPR. I would prefer if we refactor this into a `abbrev SampleM := ReaderT Config IO`, where
> 
> ```lean
> namespace Sample
> structure Config where
>     isDarwin : Bool
> 
> def SampleM := ReaderT Config IO
> 
> def Config.setup : IO Config := do
>     pure { isDarwin := true } /- do the process querying here-/
> 
> def SampleM.run (m : SampleM a) : IO a := do
>   ReaderT.run m (← Config.setup)
> 
> end Sample
> ```

@bollu Yes, fair point indeed re. the cost of `GPRIndex.rand`, and I'll take you up on your volunteer offer. :) 

But I recommend letting these changes get stress-tested in CI for a bit to see if they helped or not (I'm optimistic that they'll fix the bogus mismatch issues, but still...). If they did, then let's do that refactoring.

Another thing to note: `Config` above might as well cache other things, like the platform features we need. E.g., see https://github.com/leanprover/LNSym/blob/6d1ef32d9d253b8b7667edfaa08f38ea82da2dc1/Arm/Insts/DPSFP/Crypto_four_reg.lean#L38.

PR #100, opened 2024-08-20
feat: add hard theorems from Isabelle

bollu:
This file contains bitvector theorems that stress Isabelle automation.
We port the file to Lean for comparison.

### DIFF
--- a/Proofs/Experiments/IsabelleHard.lean
+++ b/Proofs/Experiments/IsabelleHard.lean
@@ -7,10 +7,7 @@ This file contains bitvector theorems that stress Isabelle automation.
 We port the file to Lean for comparison.
 -/
 
-import Tactics.Sym
-import Init.Data.BitVec.Folds
-import Init.Data.BitVec.Bitblast
-import Init.Data.BitVec.Lemmas
+import LeanSAT
 open BitVec
 
 /--
--- a/Proofs/Experiments/IsabelleHard.lean
+++ b/Proofs/Experiments/IsabelleHard.lean
@@ -91,7 +91,27 @@ mask_split_left
 theorem mask_split_left (x y : BitVec 64) (n m : Nat)
     (h : (x &&& (mask64 m <<< n) ||| x &&& mask64 n) = y) :
     x &&& (mask64 m <<< n) = (y &&& (mask64 m <<< n)) := by
-  bv_decide  -- sus
+  bv_decide
+
+/--
+It's highly suspect that LeanSAT can prove `mask_split_left` without
+unfolding the definition of `mask64`.
+For sanity, we prove the same theorem with
+`mask64 m` as an uninterpreted symbol called `mask64m`,
+and `mask64 n` as an uninterpreted symbol `mask64n`.
+-/
+theorem mask_split_left' (x y mask64m mask64n : BitVec 64) (n : Nat)
+    (h : (x &&& (mask64m <<< n) ||| x &&& mask64n) = y) :
+    x &&& (mask64m <<< n) = (y &&& (mask64m <<< n)) := by
+  rw [← h]
+  apply BitVec.eq_of_getLsb_eq
+  intros i
+  simp only [getLsb_and, getLsb_shiftLeft, Fin.is_lt, decide_True, Bool.true_and, getLsb_or]
+  rcases (x.getLsb i) with rfl | rfl
+  · simp
+  · simp
+    intros hi hk
+    simp [hi, hk]
 
 /--
 info: 'mask_split_left' depends on axioms: [propext, Classical.choice, Lean.ofReduceBool, Quot.sound]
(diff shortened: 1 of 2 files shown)


Proofs/Experiments/IsabelleHard.lean
```diff
@@ -0,0 +1,202 @@
...
+-/
+
+import Tactics.Sym
+import Init.Data.BitVec.Folds
+import Init.Data.BitVec.Bitblast
+import Init.Data.BitVec.Lemmas
+open BitVec
+
+/--
+`mask w n` is an w-bit word where the low `n`
```
shigoel, line 17:
```suggestion
`mask w i` is an w-bit word where the low `i`
```

Proofs/Experiments/IsabelleHard.lean
```diff
@@ -0,0 +1,202 @@
...
+mask_split_left
+∀x, y: word64. ∀n: ℕ.
+    ((x & (mask64 m << n) | x & mask64 n) = y)
+    ⟹
+    (x & (mask m64 << n) = (y & (mask64 m << n)))
+-/
+theorem mask_split_left (x y : BitVec 64) (n m : Nat)
+    (h : (x &&& (mask64 m <<< n) ||| x &&& mask64 n) = y) :
+    x &&& (mask64 m <<< n) = (y &&& (mask64 m <<< n)) := by
+  bv_decide  -- sus
```
shigoel, line 94: Why "sus"?

Proofs/Experiments/IsabelleHard.lean
```diff
@@ -0,0 +1,202 @@
...
+mask_split_right
+∀x, y: word64. ∀n: ℕ.
+    ((x & (mask64 m << n)) | (x & mask64 n)) = y)
+    ⟹
+    ((x & mask64 n) = (y & mask64 n))
+-/
+theorem mask_split_right (x y : BitVec 64) (n m : Nat)
+    (h : ((x &&& (mask64 m <<< n)) ||| (x &&& mask64 n)) = y) :
+    x &&& mask64 n = y &&& mask64 n := by
+  bv_decide -- sus
```
shigoel, line 111: Same comment: why "sus"? Let's trace `bv_decide` if this is a concern (or introduce a bug deliberately to see what the cex looks like)

Proofs/Experiments/IsabelleHard.lean
bollu:
```lean
theorem mask_split_left (x y : BitVec 64) (n m : Nat)
    (h : (x &&& (mask64 m <<< n) ||| x &&& mask64 n) = y) :
    x &&& (mask64 m <<< n) = (y &&& (mask64 m <<< n)) := by
  bv_decide  -- sus
```

I marked it as sus because I don't see how a SAT solver can prove it without unfolding what mask64 means, and IIRC LeanSAT doesn't unfold definitions. Let me understand why it succeeds.

Proofs/Experiments/IsabelleHard.lean
bollu: Same as above.

Proofs/Experiments/IsabelleHard.lean
```diff
@@ -0,0 +1,202 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Siddharth Bhat
+
+This file contains bitvector theorems that stress Isabelle automation.
+We port the file to Lean for comparison.
+-/
+
+import Tactics.Sym
```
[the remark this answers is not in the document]
bollu: Fixed. I directly import LeanSAT now.

bollu:
turns out the statement of `mask_split_left` (resp. `mask_split_right`) was stronger than necessary, and the theorem isn't about masks at all. For sanity, we prove variants where the symbols are treated as uninterpreted.

bollu:
@shigoel all comments resolved.

Proofs/Experiments/IsabelleHard.lean
```diff
@@ -0,0 +1,238 @@
...
+open BitVec
+
+/--
+`mask w n` is an w-bit word where the low `n`
+bits are `1`s and the upper bits are `0`s.
+ -/
+def mask (w : Nat) (i : Nat) : BitVec w :=
+  match i with
+  | 0 => 0#w
+  | i' + 1 => 1#w <<< i' ||| mask w i'
```
tobiasgrosser, line 20: Rather than a recursive match, can you just make this left shift of allOnes and then negate this? This seem to preserver more of the structure.

shigoel:
@bollu Shall we close this PR?

bollu:
Yep, closing.

PR #104, opened 2024-08-21
chore: make cosim quicker on default build

bollu:
I realized that I was never really running cosim locally because it took too long --- in an effort to remedy the situation, I set `make` (the default `make all`) rule to run `quick-cosim`, which sets `--num-tests=3`. While this is still a tad too long for my taste, it is, at the very least, tolerable, so I actually run cosim locally now rather than relying on CI (: 

### License:

### DIFF
--- a/Main.lean
+++ b/Main.lean
@@ -20,7 +20,7 @@ def find_index (e : String) (xs : List String) : Option Nat :=
   find_index_aux e xs 0
 
 -- Default number of tests/instruction
-def default_numTests : Nat := 20
+def default_numTests : Nat := 3
 
 def numTests_option (args : List String) : IO Nat :=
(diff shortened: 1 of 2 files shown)


shigoel:
Instead of adding a new target, I recommend simply setting the default `NUM_TESTS` to 3. Anyway in CI, we run [`make cosim NUM_TESTS=100`](https://github.com/leanprover/LNSym/blob/5f90bf972968a0414dde89763de8da28a9274b8b/.github/workflows/makefile.yml#L54) which does the heavyweight conformance testing.

bollu:
@shigoel updated PR to set `default_numTests` to `3`.

shigoel (approved):
LGTM, thanks!

PR #113, opened 2024-08-22
feat: enable fuel for CSE, cleanup logging for CSE. 

bollu:
The example is pretty brutal, so I fully believe the naive implementation does not scale! Thanks for the call out @shigoel. I'll be using this example to fix scaling issues in CSE.

PR #114, opened 2024-08-22
chore: checkpoint tactic state before  in CSE and restore on failure [1/4?]

bollu:
This fixes the bug reported by @shigoel, where we would get a `no goals to be solved` failure after the `cse` tactic was run on all hypotheses --- we were incorrectly backtracking, which is fixed now. 

The next PRs will improve scaling issues by the following:
1. adding a custom amount of fuel that will allow the CSE tactic to stop gracefully.
2. Not CSEing small objects such as `Nat` literals which are ubiquitous, as well as pointless to CSE.

We will also make improvements to debug logging in a separate commit. 
Thus, we anticipate this patch set to have four patches, which will be peeled from https://github.com/leanprover/LNSym/pull/113

### DIFF
--- a/Proofs/SHA512/SHA512Sym.lean
+++ b/Proofs/SHA512/SHA512Sym.lean
@@ -47,7 +47,7 @@ abbrev ktbl_addr : BitVec 64 := 0x1b4300#64
 -- set_option profiler true in
 -- set_option profiler.threshold 1 in
 -- set_option pp.deepTerms false in
-set_option maxHeartbeats 9999999 in
+set_option maxHeartbeats 9999999 in -- To be fixed by https://github.com/leanprover/LNSym/pull/113
 theorem sha512_block_armv8_1block (s0 sf : ArmState)
   -- (FIXME) Ignore the `stp` instruction for now.
   (h_s0_pc : read_pc s0 = 0x1264c4#64)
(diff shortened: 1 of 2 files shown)


Proofs/SHA512/SHA512Sym.lean
```diff
@@ -47,6 +47,7 @@ abbrev ktbl_addr : BitVec 64 := 0x1b4300#64
 -- set_option profiler true in
 -- set_option profiler.threshold 1 in
 -- set_option pp.deepTerms false in
+set_option maxHeartbeats 9999999 in
```
shigoel, line 50: I'm hoping this is stopgap, right? Adding a fuel arg. and treating literals as atomic should help?

shigoel, line 50: I'm hoping this is stopgap, right? Adding a fuel arg. and treating literals as atomic should help?

If so, please add a `FIXME` or something there to indicate it'd go away soon.

bollu: That's right. In the final implementation I will upstream (https://github.com/leanprover/LNSym/pull/113/files#diff-51d27053b63053ca91adbfef85c82a2d7fe480198ef4b1d2cbdda5934e393de0R50), we won't have a `maxHeartbeats`.

bollu:

shigoel (approved):

PR #119, opened 2024-08-26
Total Correctness Framework

shigoel:
### Description:

General method for proving total correctness based on proof strategies in the paper
https://link.springer.com/article/10.1007/s10817-008-9098-1

### Testing:

`make all` succeeded on my M3.

### License:


Correctness/Correctness.lean
```diff
@@ -21,44 +24,68 @@ open Spec' (cut assert)
...
         ⟨n, Nat.le_refl _, he, hpost⟩
     find 0 (ini _ hp) (Nat.zero_le _)
 
+----------------------------------------------------------------------
+
+-- Prove partial correctness using inductive assertions using the
+-- functions `csteps` (to characterize the number of steps to reach
+-- the next cutpoint) and `nextc` (to characterize the next cutpoint
+-- state). Note that the function `csteps` is partial: if no cutpoint
+-- is reachable from `s`, then the recursion does not terminate.
```
alexkeizer, line 87:
```suggestion
/-!
Prove partial correctness using inductive assertions using the
functions `csteps` (to characterize the number of steps to reach
the next cutpoint) and `nextc` (to characterize the next cutpoint
state). Note that the function `csteps` is partial: if no cutpoint
is reachable from `s`, then the recursion does not terminate.
-/
```

This kind of comment will show up in docs

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
         ⟨n, Nat.le_refl _, hexit, hpost⟩
     find 0 (v1 s0 hp) (Nat.zero_le ..)
 
+----------------------------------------------------------------------
+
+-- Prove partial correctness using inductive assertions using the
+-- function `cassert` that checks if `assert` holds if at a cutpoint
+-- state, or else recurs until it hits one. This function is also
+-- partial: if no cutpoint is reachable, the recursion doesn't
+-- terminate.
```
alexkeizer, line 224:
```suggestion
/-!
Prove partial correctness using inductive assertions using the
function `cassert` that checks if `assert` holds if at a cutpoint
state, or else recurs until it hits one. This function is also
partial: if no cutpoint is reachable, the recursion doesn't
terminate.
-/
```

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+    (fun (si, i) =>
+        if cut si then
+          .inl (i, assert s0 si)
+        else
+          .inr (next si, i + 1))
+    (si, i)
+
+theorem cassert_eq [Sys σ] [Spec' σ] (s0 si : σ) (i : Nat)
+        : cassert s0 si i = if cut si then (i, assert s0 si)
+                       else cassert s0 (next si) (i + 1) := by
```
alexkeizer, line 243:
```suggestion
theorem cassert_eq [Sys σ] [Spec' σ] (s0 si : σ) (i : Nat) :
    cassert s0 si i 
    = if cut si then (i, assert s0 si)
       else cassert s0 (next si) (i + 1) := by
```

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+  unfold cassert
+  conv => lhs; rw [iterate_eq]
+  by_cases cut si <;> simp only [Bool.false_eq_true, ↓reduceIte, *]
+  done
+
+theorem cassert_cut [Sys σ] [Spec' σ] {s0 si : σ} (h : cut si) (i : Nat) :
+  (cassert s0 si i).fst = i ∧
+  (cassert s0 si i).snd = assert s0 si := by
```
alexkeizer, line 251:
```suggestion
    (cassert s0 si i).fst = i ∧
    (cassert s0 si i).snd = assert s0 si := by
```
Not sure whether we want to keep following the style we currently have (which has the statement and proof at the same indentation), or slowly migrate toward standard Lean style (which has the statement indented one more level than the proof, as in my suggestion)

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+-/
+theorem partial_correctness_from_assertions [Sys σ] [Spec' σ]
+    (v1 : ∀ s0 : σ, pre s0 → assert s0 s0)
+    -- (FIXME) Is it possible to remove v2 and combine v3 and v4 into
+    -- a single verification condition as follows?
+    -- As @bollu noted, this has the benefit of not having `post` be a
+    -- callee of `assert`.
+    --
+    --  ∀ s0 si : σ, assert s0 si → (cassert s0 (run si 1) 0).snd ∨
+    --                              (exit si → post s0 si)
```
alexkeizer, line 383: Shouldn't it be `exit si /\ post s0 si`? Otherwise, the condition will be trivial for non-exit states

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+
+----------------------------------------------------------------------
+
+-- Method to prove Termination
+
+-- We follow the same formalism for the termination proof like we did for
+-- `partial_correctness_from_assertions`; i.e., we reuse `cassert`. The goal is
+-- to be able to prove partial correctness and termination at the same time.
+
+/-
```
alexkeizer, line 427:
```suggestion
/--
```

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+  conv => lhs; rw [iterate_eq]
+  by_cases cut sn <;> simp only [Bool.false_eq_true, ↓reduceIte, *]
+  done
+
+/-
+If a cutpoint is reachable from `si` in some `n` steps (i.e.,`cut (run si n)`)
+and `(cassert s0 si i).snd`, then we know that `assert` holds from `si` after
+`((cassert s0 si i).fst - i)` steps.
+-/
+theorem cassert_when_cut_exists [Sys σ] [Spec' σ] (s0 si : σ) (n i : Nat)
```
alexkeizer, line 457:
```suggestion
/--
If a cutpoint is reachable from `si` in some `n` steps (i.e.,`cut (run si n)`)
and `(cassert s0 si i).snd`, then we know that `assert` holds from `si` after
`((cassert s0 si i).fst - i)` steps.
-/
theorem cassert_when_cut_exists [Sys σ] [Spec' σ] (s0 si : σ) (n i : Nat)
```

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+  have := @term_helper_aux σ _ _ s0 si n h_cut v2' h_not_exit h_assert
+  apply Exists.intro n
+  exact ⟨this, h_rank⟩
+  done
+
+/-
+Termination holds if `v1` and `v2` are true. Note the similarities with
+`partial_correctness_from_assertions`: `v1` of both these theorems match, and
+`v2` of this theorem is very similar to `v4` of the other.
+-/
```
alexkeizer, line 578:
```suggestion
/--
Termination holds if `v1` and `v2` are true. Note the similarities with
`partial_correctness_from_assertions`: `v1` of both these theorems match, and
`v2` of this theorem is very similar to `v4` of the other.
-/
```

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+                 h_exit_assump (h_term_helper.left) (by rfl)
+    done
+
+----------------------------------------------------------------------
+
+/-
+A method to prove total correctness (`PartialCorrectness` + `Termination`) in
+one go.
+-/
+protected theorem by_the_method [Sys σ] [Spec' σ] (rank : σ → Nat)
```
alexkeizer, line 622:
```suggestion
/--
A method to prove total correctness (`PartialCorrectness` + `Termination`) in
one go.
-/
protected theorem by_the_method [Sys σ] [Spec' σ] (rank : σ → Nat)
```

alexkeizer (commented):
I've mostly got comments about style and documentation, none of which I'd consider particularly blocking

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+  done
+
+theorem cassert_not_cut [Sys σ] [Spec' σ] {s0 si : σ} (h₁ : ¬ cut si)
+  (h₂ : (cassert s0 (next si) (i+1)).fst = j) :
+  (cassert s0 si i).fst = j ∧
+  (cassert s0 si i).snd = (cassert s0 (next si) (i + 1)).snd := by
+  rw [cassert_eq]
+  simp only [h₁, Bool.false_eq_true, ↓reduceIte, and_true]
+  assumption
+  done
```
bollu, line 263:
```suggestion
```

No need for `done` :)

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+-- to be able to prove partial correctness and termination at the same time.
+
+/-
+`rank_decreases rank si sn i`: if this function terminates, it returns a pair
+whose first element is a `Nat` that is `(i + the number of steps simulated
+from si to reach the next cutpoint)`, and the second element is a
+`Prop` that checks whether the `rank` of the next cutpoint state is strictly
+less than the `rank` of `si`.
+-/
+noncomputable def rank_decreases [Sys σ] [Spec' σ] (rank : σ → Nat) (si sn : σ) (i : Nat)
```
bollu, line 434:
```suggestion
less than the `rank` of `si`.

We may eventually want to generalize from `Nat` to any `WellFounded` (using https://github.com/leanprover/lean4/blob/a5162ca7489bfdbf1a2851cffd8fdcca9d2b9b56/src/Init/WF.lean#L46-L54). This is particularly useful for nested loops, whose termination arguments often involve tuples of naturals with lexicographic ordering (i.e. https://en.wikipedia.org/wiki/Loop_dependence_analysis)
-/
noncomputable def rank_decreases [Sys σ] [Spec' σ] (rank : σ → Nat) (si sn : σ) (i : Nat)
```

bollu (commented):
LGTM, I added some notes of generalizing the rank function to ordinals.

Correctness/Correctness.lean
shigoel: But it's so satisfying!

Correctness/Correctness.lean
shigoel: Thanks!

Correctness/Correctness.lean
shigoel: Like a flourish!

Correctness/Correctness.lean
```diff
@@ -175,4 +215,431 @@ theorem partial_correctness_from_verification_conditions [Sys σ] [Spec' σ]
...
+  loop si 0 (Nat.zero_le ..) rfl
+
+/--
+Prove partial correctness from inductive assertions using `cassert`
+function.
+
+We use `s0`, `si`, and `sf` to refer to initial, intermediate, and
+final (exit) states respectively.
+
+This is more convenient to use that
```
shigoel, line 376:
```suggestion
This is more convenient to use than
```

Correctness/Correctness.lean
shigoel: Yes, thanks for noticing. What's written is definitely wrong, and I haven't figured out the right form yet (hence the TODO) :) 
Let's take up what you suggest:
`a -> c or (e and p)` (I'm using `a` for `assert`, `c` for cassert`, `e` for `exit`, and `p` for `post`).

Suppose `e` is `true`. In that case, we have:
`a -> c or p`

But we really want `p` to hold at an exit state, and this formulation too does not enforce that.

We can combine `v3` and `v4` to be something like:
`a -> (e or c) and (not e or p)`
but I need to work out how to remove `v2`.

PR #122, opened 2024-08-26
chore: Memory Tactic now solves in a loop and throws error if failed to make progress 

bollu:
This removed the need for repeated calls for `simp_mem`, and also uses the information in the state to throw an error if the tactic was unable to make *any* progress whatsoever. We mimic the `simp` API, and allow this to be changed by setting `failIfUnchanged? := false`.


Proofs/Experiments/MemoryAliasing.lean
```diff
@@ -159,7 +159,7 @@ theorem mem_automation_test_4
       (write_mem_bytes 48 src_addr val s0)) =
    val.extractLsBytes 1 10 := by
   simp only [memory_rules]
-  simp_mem; simp_mem -- TODO: repeat on change.
+  simp_mem; -- TODO: repeat on change.
```
shigoel, line 162:
```suggestion
  simp_mem
```
TODO resolved, right?

Arm/Memory/SeparateAutomation.lean
```diff
@@ -697,99 +704,130 @@ partial def SimpMemM.simplifyExpr (e : Expr) (hyps : Array Hypothesis) : SimpMem
...
+
+      let changed? ← withTraceNode m!"Performing Rewrite At Main Goal" do
         SimpMemM.simplifyGoal (← getMainGoal) foundHyps
+      madeProgress? := madeProgress? || changed?
 
+      if changed?
+      then continue
+        else if !madeProgress? && (← getConfig).failIfUnchanged then
+          throwError "{crossEmoji} simp_mem failed to make progress."
+        break
```
shigoel, line 830: I find the indentation confusing. Mind lining the if/then/elses up for me? Thx!

bollu: I cleaned up the control flow instead :) Hope the new version reads better.

shigoel (approved):
Thanks, @bollu!

PR #126, opened 2024-08-27
feat: add automation support for pairwise memory separation

bollu:
This is stacked on top of #123 , since we want to state this, for a given `ms : List MemoryRegion`, `ms.pairwise MemoryRegion.Separate`. This, of course, needs us to have the `MemoryRegion` abstraction in the first place.


shigoel:
@bollu  Conflicts here.

bollu:
@shigoel conflicts addressed!

Arm/Memory/Separate.lean
```diff
@@ -328,6 +328,12 @@ theorem BitVec.not_le_eq_lt {a b : BitVec w₁} : (¬ (a ≤ b)) ↔ b < a := by
...
   omega
 
+theorem mem_separate'_comm (h : mem_separate' a an b bn) :
+    mem_separate' b bn a an := by
+  have := h.omega_def
+  apply mem_separate'.of_omega
+  omega
+
 /-#
 This is a theorem we ought to prove, which establishes the equivalence
```
shigoel, line 338: Just got reminded of this TODO theorem: `mem_separate_of_mem_separate'` -- let's add an issue to keep track of this.

Arm/Memory/Separate.lean
```diff
@@ -560,4 +566,53 @@ theorem Memory.read_bytes_eq_extractLsBytes_sub_of_mem_subset'
       bv_omega
     · simp only [h, ↓reduceIte]
 
+/-- A region of memory, given by (base pointer, length) -/
+abbrev Memory.Region := BitVec 64 × Nat
```
shigoel, line 570: Glad you chose this name, I'm partial to it somehow :)

Arm/Memory/Separate.lean
```diff
@@ -560,4 +566,53 @@ theorem Memory.read_bytes_eq_extractLsBytes_sub_of_mem_subset'
...
+  (hij : i ≠ j)
+  (a b : Memory.Region)
+  (ha : mems.get? i = some a) (hb : mems.get? j = some b) :
+    mem_separate' a.fst a.snd b.fst b.snd := by
+  induction h generalizing a b i j
+  case nil => simp only [List.get?_eq_getElem?, List.getElem?_nil, reduceCtorEq] at ha
+  case cons x xs ihx _ihxs ihxs' =>
+    simp only [List.get?_eq_getElem?] at ha hb
+    rcases i with rfl | i'
+    · simp at ha
```
shigoel, line 597: `simp only` please. Yes, I know I'm paranoid. :)

Proofs/Experiments/MemoryAliasing.lean
```diff
@@ -10,8 +10,8 @@ import Arm.Memory.MemoryProofs
 import Arm.BitVec
 import Arm.Memory.SeparateAutomation
 
-set_option trace.simp_mem true
-set_option trace.simp_mem.info true
+-- set_option trace.simp_mem true
```
shigoel, line 13: Thank you!

Proofs/Experiments/MemoryAliasing.lean
```diff
@@ -366,17 +382,37 @@ theorem mem_separate_move_of_lt_of_le  (h : mem_separate' a an b bn)
...
+  theorem pairwise_subset (h : Memory.Region.pairwiseSeparate [⟨a, 100⟩, ⟨b, 200⟩, ⟨c, 300⟩, ⟨d, 400⟩]) :
+    mem_separate' a 80 b 100 := by
+    simp_mem
+
+end PairwiseSeparate
+
 namespace MemOptions
 
+set_option trace.simp_mem true
+set_option trace.simp_mem.info true
```
shigoel, line 403: Comment for less verbose `make` output.

shigoel, line 403:
```suggestion
-- set_option trace.simp_mem true
-- set_option trace.simp_mem.info true
```

shigoel: Oh wait, do you need these to be turned on for the `#guard_msgs` below?

Arm/Memory/SeparateAutomation.lean
```diff
@@ -308,19 +361,36 @@ def consumeRewriteFuel : SimpMemM Unit :=
...
-def introDef (name : String) (hdefVal : Expr) : SimpMemM FVarId  := do
+def simpAndIntroDef (name : String) (hdefVal : Expr) : SimpMemM FVarId  := do
   SimpMemM.withMainContext do
+
     let name ← mkFreshUserName <| .mkSimple name
     let goal ← getMainGoal
     let hdefTy ← inferType hdefVal
+
+    /- Simp to gain some more juice out of the defn.. -/
+    let (simpCtx, simprocs) ← LNSymSimpContext
```
shigoel, line 388: This is a heavy `simp`, what with `state_simp_rules`, `bitvec_rules`, and `minimal_theory` in there. Any pitfalls?

Proofs/Experiments/MemoryAliasing.lean
```diff
@@ -159,7 +175,7 @@ theorem mem_automation_test_4
       (write_mem_bytes 48 src_addr val s0)) =
    val.extractLsBytes 1 10 := by
   simp only [memory_rules]
-  simp_mem; -- TODO: repeat on change.
+  simp_mem -- TODO: repeat on change.
```
shigoel, line 178:
```suggestion
  simp_mem
```
The TODO got done previously, right?

Arm/Memory/SeparateAutomation.lean
```diff
@@ -403,6 +473,35 @@ def MemSeparateProp.ofExpr? (e : Expr) : Option MemSeparateProp :=
...
   | _ => none
 
+/-- info: Prod.mk.{u, v} {α : Type u} {β : Type v} (fst : α) (snd : β) : α × β -/
+#guard_msgs in #check Prod.mk
+
+/-- info: List.cons.{u} {α : Type u} (head : α) (tail : List α) : List α -/
+#guard_msgs in #check List.cons
+
+/-- info: List.nil.{u} {α : Type u} : List α -/
+#guard_msgs in #check List.nil
```
shigoel, line 483:
```suggestion
```
We don't need these at this point, I think. It's fairly clear what's going on at `MemPairwiseSeparateProof.ofExpr?`.

Arm/Memory/SeparateAutomation.lean
```diff
@@ -482,10 +585,74 @@ def MemSeparateProof.omega_def (h : MemSeparateProof e) : Expr :=
...
     trace[simp_mem.info]  "{h}: added omega fact ({h.omega_def})"
     return args.push (Expr.fvar fvar)
 
+
+
+/-- info: Ne.{u} {α : Sort u} (a b : α) : Prop -/
+#guard_msgs in #check Ne
+
+/-- info: List.get?.{u} {α : Type u} (as : List α) (i : Nat) : Option α -/
+#guard_msgs in #check List.get?
```
shigoel, line 599:
```suggestion
```

Arm/Memory/SeparateAutomation.lean
```diff
@@ -482,10 +585,74 @@ def MemSeparateProof.omega_def (h : MemSeparateProof e) : Expr :=
...
+  let lhs ← mkAppOptM ``List.get? #[.none, mems.getMemSpanListExpr, mkNatLit i]
+  let rhs ← mkSome MemSpanExpr.toTypeExpr a.toExpr
+  mkEq lhs rhs
+
+/--
+info: Memory.Region.separate'_of_pairwiseSeprate_of_mem_of_mem {mems : List Memory.Region}
+  (h : Memory.Region.pairwiseSeparate mems) (i j : Nat) (hij : i ≠ j) (a b : Memory.Region) (ha : mems.get? i = some a)
+  (hb : mems.get? j = some b) : mem_separate' a.fst a.snd b.fst b.snd
+-/
+#guard_msgs in #check Memory.Region.separate'_of_pairwiseSeprate_of_mem_of_mem
```
shigoel, line 612: Unlike other `#guard_msgs/#check` blocks, this is a nice reminder because `Memory.Region.separate'_of_pairwiseSeprate_of_mem_of_mem` is nontrivial.

Arm/Memory/SeparateAutomation.lean
```diff
@@ -482,10 +585,74 @@ def MemSeparateProof.omega_def (h : MemSeparateProof e) : Expr :=
...
+      b.toExpr,
+      haVal,
+      hbVal]
+  return ⟨h⟩
+/--
+Currently, if the list is syntacticaly of the form [x1, ..., xn],
+ we create hypotheses of the form `mem_separate' xi xj` for all i, j..
+This can be generalized to pairwise separation given hypotheses x ∈ xs, x' ∈ xs.
+-/
+def MemPairwiseSeparateProof.addOmegaFacts (h : MemPairwiseSeparateProof e) (args : Array Expr) :
```
shigoel, line 642: This looks like a lot of omega facts, and maybe that's fine performance-wise, but I'm still nervous.

Can we think about getting these facts smartly/lazily from the context instead? Lot more work, plus we need to make sure we need to be that smart. Thoughts?

shigoel:
Left some comments, looks good in general.

We need more tests for the new pairwise construct (in a PR in the very near future is okay!) -- let's start with changing theorems in the [SHA512 example](https://github.com/leanprover/LNSym/blob/main/Proofs/Experiments/SHA512MemoryAliasing.lean) to use `pairwise`, and then we'll go from there?

Arm/Memory/Separate.lean
bollu: done: https://github.com/leanprover/LNSym/issues/137

Arm/Memory/SeparateAutomation.lean
bollu: @shigoel I thought about this a bit, and unfortunately, I convinced myself that there's no complete heuristic for this, so we are forced to know all $$O(n^2)$$ facts in the worst case. Intuitively, we way need to establish something based on knowing the mutual exclusion of *all* memory regions. 

I would like to keep a watch on this, though, and see how it scales as we get larger examples.

Proofs/Experiments/MemoryAliasing.lean
bollu: Still, I scoped it with `set_option trace.simp_mem true in`...

Arm/Memory/SeparateAutomation.lean
bollu: @shigoel I prefer to leave them in, since it at least helps me sanity check what's going on --- code is read a lot more than it's written!

bollu:

Arm/Memory/SeparateAutomation.lean
shigoel: Agree with the intuition, and this works for now.

I was suggesting a lazy pick-a-point heuristic for the future (if need be, and we don't know that until we try bigger examples, I agree): keep picking a separateness fact from the list of memory regions until the goal can't be simplified any further, in which case we throw all the facts in the context. This could work in the common case; very rarely do we need to establish something from the mutual exclusion of _all_ the regions.

But this could be an awful heuristic because of the back-and-forth between picking a fact and attempted simplification; we could build a way to turn this heuristic off. I don't know. Anyway, just a thought for now.

Arm/Memory/SeparateAutomation.lean
shigoel: @bollu ?

Arm/Memory/Separate.lean
shigoel: @bollu ?

shigoel:
@bollu I see that a couple comments weren't addressed. Tagged you there.

Arm/Memory/SeparateAutomation.lean
bollu: bitvec_rules, minimal_theory, `state_value`.

bollu: Addressed!

bollu:
@shigoel All comments (hopefully) addressed.

Arm/Memory/SeparateAutomation.lean
```diff
@@ -385,9 +388,20 @@ def simpAndIntroDef (name : String) (hdefVal : Expr) : SimpMemM FVarId  := do
...
+        | throwError m!"[simp_mem] Internal error: simp attribute {a} not found!"
+      simpTheorems := simpTheorems.push (← ext.getTheorems)
+
+    -- unfold `state_value.
+    simpTheorems := simpTheorems.push <| ← ({} : SimpTheorems).addDeclToUnfold `state_value
+    let simpCtx : Simp.Context := { 
+      simpTheorems,
+      config := { decide := true, failIfUnchanged := false },
+      congrTheorems := (← Meta.getSimpCongrTheorems)
+    }
```
shigoel, line 403:
```suggestion
LNSymSimpContext 
   (config := { decide := false, failIfUnchanged := false }) 
   (simp_attrs := #[`minimal_theory, `bitvec_rules])
   (decls_to_unfold := #[state_value])
```

@bollu I believe `LNSymSimpContext` can work here. Fair warning: haven't tried out the suggestion, excuse typos, etc.

shigoel: Also, open to making `LNSymSimpContext` nicer (separate issue).

bollu: @shigoel I think I would prefer to keep it as-is, because it's hard to see what `LNSymSimpContext` builds unless one goes and reads its code. Since this code is meant to be performant, I would rather have the building of the simp-set inline, so it's easy to audit.

bollu:
@shigoel happy to have this merged, as I prefer keeping the current around to build the `SimpContext` around.

Arm/Memory/SeparateAutomation.lean
shigoel: Weren't you using `LNSymSimpContext` before though?
https://github.com/leanprover/LNSym/pull/126/commits/dd24b6dd0f2c6d7fd8ab0cde42ccd1e75b0838df

Not a big deal, but I want to make sure I understand what the reasons are.

bollu: Indeed, I was reusing `LNSymSimpContext` to make sure that if `LNSymSimpContext` got more powerful, I would get the same power for free :) But thinking this through, I found your argument for why this should *not* be  expensive to be compelling. Therefore, I'd prefer to build a custom simp-set that's tuned down, and to add power to it as necessary. 

From an engineering perspective, I'm advocating for [WET / write everything twice] (https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) instead of DRY. 😉

shigoel: Gotcha, though I would note that the particular invocation of `LNSymSimpContext` is equivalent to what you wrote inline.

PR #128, opened 2024-08-27
feat: prove specification of Max program that manipulates memory via new VCG

bollu:
Hopefully supercedes https://github.com/leanprover/LNSym/pull/125, builds on top of #119


bollu:
Superceded by https://github.com/leanprover/LNSym/pull/136, but there is useful theory on the branch, so I'm keeping the branch around for now, so I can peel the useful theory out

PR #129, opened 2024-08-27
feat: efficient axiomatization of effects

alexkeizer:
### Description:

We've built a new datastructure `AxEffects`, which stores a map from `StateField` to two `Expr`s: the current value of that field, and a proof that `r <field> <currentState> = <value>`, together with other assorted proofs related to axiomatic effects.

We then incorporate this `AxEffects` into `sym_n`, replacing and superseding `intro_fetch_decode`. This has two benefits: `AxEffects` is more complete, and it is *much* faster. Fully simulating `gcm_gmult_v8_program` (assuming step theorems have been pre-generated) used to take around 3.5 seconds, with this PR it takes only 0.5 seconds: a 7x speedup!

- In the process, we've proven that `Abs` correctly implements its spec
- No longer is a `CheckSPAlignment` goal added by default. Instead, we only add `CheckSPAlignment` goals if the initial state was known to be aligned, and there is some write to the SP of which alignment could not be automatically proven. For now, this goal is added eagerly, without trying to determine if this fact is actually needed for further simulation. Trying to be smarter about this is left for future PRs.

### Testing:

`make all` works locally

### License:


alexkeizer:
@bollu 

I've got a memory-related theorem left to prove:
```lean
theorem read_mem_bytes_write_mem_bytes_of_read_mem_eq
    (h : ∀ n addr, read_mem_bytes n addr s₁ = read_mem_bytes n addr s₂)
    (n₂ addr₂ val n₁ addr₁) :
    read_mem_bytes n₁ addr₁ (write_mem_bytes n₂ addr₂ val s₁)
    = read_mem_bytes n₁ addr₁ (write_mem_bytes n₂ addr₂ val s₂) := by
```
Informally, this should be true: either we read a byte from `[addr₂...n₂]`, which has the same value written to it on both sides, or we read a byte from outside that range, in which case `h` applies.

Any ideas how to convince Lean of this?

bollu:
@alexkeizer and I paired on the theorem :)

Arm/State.lean
```diff
@@ -478,6 +482,45 @@ def write_err (v : StateError) (s : ArmState) : ArmState :=
...
+
+def toFin : StateField → Fin 70
+  | .GPR x    => x.toFin.castAdd 38
+  | .SFP x    => (x.toFin.addNat 32).castAdd 6
+  | .PC       => 64
+  | .FLAG .N  => 65
+  | .FLAG .Z  => 66
+  | .FLAG .C  => 67
+  | .FLAG .V  => 68
+  | .ERR      => 69
```
bollu, line 498:
```suggestion
/-- Might eventually be used to maintain `O(1)` access by using `StateField.toFin` as keys to an array access. -/
def toFin : StateField → Fin 70
  | .GPR x    => x.toFin.castAdd 38
  | .SFP x    => (x.toFin.addNat 32).castAdd 6
  | .PC       => 64
  | .FLAG .N  => 65
  | .FLAG .Z  => 66
  | .FLAG .C  => 67
  | .FLAG .V  => 68
  | .ERR      => 69
```

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,563 @@
...
+That is, the given `state` is assigned to be both the initial and the current
+state -/
+def initial (state : Expr) : AxEffects where
+  initialState      := state
+  currentState      := state
+  fields            := .empty
+  nonEffectProof    :=
+    mkLambda `f .default (mkConst ``StateField) <|
+      let rf := mkApp2 (mkConst ``r) (.bvar 0) state
+      mkApp2 (.const ``Eq.refl [1]) (.const ``ArmState []) rf
```
bollu, line 92:
```suggestion
      mkApp2 (.const ``Eq.refl [1]) (.const ``ArmState) rf
```

bollu, line 92: nit: use the `mkEqArmState` helper.

alexkeizer: This actually needs a `mkEqReflArmState` helper, so I added that

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,562 @@
...
+      mkAppN (mkConst ``CheckSPAlignment_of_r_sp_eq)
+        #[eff.initialState, eff.currentState, proof, spAlignment]
+    trace[Tactic.sym] "constructed stackAlignmentProof: {stackAlignmentProof?}"
+    return some { eff with stackAlignmentProof? }
+
+/-! ## Composition -/
+
+/- TODO: write a function that combines two effects `left` and `right`,
+where `left.initialState = right.currentState`.
+That is, compose the effect of "`left` after `right`" -/
```
alexkeizer, line 492: This TODO is left for a future PR

Arm/BitVec.lean
```diff
@@ -555,6 +555,11 @@ protected theorem extract_lsb_of_zeroExtend (x : BitVec n) (h : j < i) :
   simp_all
   omega
 
+@[bitvec_rules, simp]
```
shigoel, line 558: Is the `simp` tag necessary?

Arm/Insts/Common.lean
```diff
@@ -144,6 +144,18 @@ theorem CheckSPAlignment_AddWithCarry_64_4 (st : ArmState) (y : BitVec 64) (carr
   simp_all only [CheckSPAlignment, read_gpr, zeroExtend_eq, Nat.sub_zero, add_eq,
     Aligned_AddWithCarry_64_4]
 
+theorem CheckSPAlignment_of_r_sp_eq {s s' : ArmState}
```
shigoel, line 147: Needs a `state_simp_rules` tag? Also for the theorem below?

Arm/State.lean
```diff
@@ -255,7 +255,11 @@ inductive StateField where
   | PC     : StateField
   | FLAG   : PFlag → StateField
   | ERR    : StateField
-deriving DecidableEq, Repr
+deriving DecidableEq, Repr, Hashable
+
+/-- general purpose register `x31` is used as stack pointer -/
+@[state_simp_rules]
+abbrev StateField.SP := StateField.GPR 31#5
```
shigoel, line 262: Nice! We ought to add such abbreviations for other registers as well in a future PR (and perhaps something like this is in one of the pending PRs). I know @bollu has thoughts re. notation.

Arm/State.lean
```diff
@@ -478,6 +482,47 @@ def write_err (v : StateError) (s : ArmState) : ArmState :=
 
 end Accessor_updater_functions
 
+----------------------------------------------------------------------
+---- `StateField` is equivalent to `Fin (32 + 32 + 1 + 4 + 1)` -------
```
shigoel, line 486: Belongs right above `toFin`.

Arm/State.lean
```diff
@@ -478,6 +482,47 @@ def write_err (v : StateError) (s : ArmState) : ArmState :=
...
+
+namespace StateField
```
shigoel, line 488: Let's move this section below the definition of `StateField` for data locality reasons.

Arm/State.lean
```diff
@@ -645,6 +690,13 @@ theorem read_mem_bytes_of_w :
     rw [n_ih]
   done
 
+theorem read_mem_bytes_w_of_read_mem_eq
```
shigoel, line 693: Add `state_simp_rules` tag.

Arm/State.lean
```diff
@@ -645,6 +690,13 @@ theorem read_mem_bytes_of_w :
...
+    (h : ∀ n addr, read_mem_bytes n addr s₁ = read_mem_bytes n addr s₂)
+    (fld val n₁ addr₁) :
```
shigoel, line 695: I'm curious why you chose to have explicit args here.

Arm/State.lean
shigoel: BTW, I know this ruleset is aggressive at the moment, but it's good to have all the relevant theorems in there in case we want to revert to doing symbolic simulation manually for whatever reason.

Arm/State.lean
```diff
@@ -1022,3 +1074,32 @@ by_cases h : ix < base
   · simp only [h₂, ↓reduceIte, BitVec.getLsb_extractLsByte]
 
 end Memory
+
+/-! ## Helper lemma for `AxEffects` -/
+
+theorem Memory.eq_of_read_mem_bytes_eq {m₁ m₂ : Memory}
```
shigoel, line 1080: Ah, this reminds me: we have a file [`StateEq.lean`](https://github.com/leanprover/LNSym/blob/main/Arm/StateEq.lean) which has such eq lemmas. Unfortunately, I didn't imported it anywhere, so it didn't keep up with changes to LNSym.

I'll create an issue to update that file.

Arm/State.lean
```diff
@@ -1022,3 +1074,32 @@ by_cases h : ix < base
   · simp only [h₂, ↓reduceIte, BitVec.getLsb_extractLsByte]
 
 end Memory
+
+/-! ## Helper lemma for `AxEffects` -/
```
shigoel, line 1078: BTW, `state_simp_rules` tag for these theorems too.

Proofs/Experiments/Abs.lean
```diff
@@ -29,15 +38,29 @@ theorem correct
   (h_s0_program : s0.program = program)
   (h_s0_err : read_err s0 = StateError.None)
   (h_s0_sp : CheckSPAlignment s0)
-  (h_run : sf = run program.length s0) :
+  (h_run : sf = run (program.length - 1) s0) :
```
shigoel, line 41:
```suggestion
  -- (FIXME) We do not account for the effects of `ret` yet, so 
  -- we simulate `program.length - 1` number of instructions only.
  (h_run : sf = run (program.length - 1) s0) :
```

shigoel: Though... I believe for this particular program, since we don't see any SP action, we could try to simulate `ret` too. Want to try, @alexkeizer?

Tactics/Attr.lean
```diff
@@ -8,3 +8,8 @@ initialize
 
   -- enable tracing for `sym_n` tactic and related components
   registerTraceClass `Tactic.sym
+
+  registerOption `Tactic.sym.debug {
+    defValue := true
+    descr := "enable/disable extra validation of `sym_n` internal state"
```
shigoel, line 14: "validation" is an interesting but perhaps overloaded word that could also mean "proof". Would "inspection" be more suitable, or am I missing something?

Tests/Tactics/Sym.lean
```diff
@@ -17,6 +20,29 @@ namespace GCMGmultV8Program
...
+    (h_program : s0.program = program)
+    (h_run : sf = run 1 s0)
+    (h_err : read_err s0 = .None)
+    (h_pc : read_pc s0 = 0x4005b8#64)
+    (h_sp : CheckSPAlignment s0)
+    (h_s0_x0 : r (.GPR 0#5) s0 = x)
+    (h_s0_x31 : r (.GPR 31#5) s0 = y) :
+    read_mem_bytes 4 (y + 12#64) s0 = (x.zeroExtend 32) := by
+  sym_n 1
+  sorry
```
shigoel, line 46: @bollu and @alexkeizer: Excellent example to kick-start thinking about to get `simp_mem` and `sym_n` get along great together!

Tactics/Attr.lean
alexkeizer: This option does not control tracing, it causes `AxEffects.validate` to be called at certain points, which will typecheck the various expressions stored in it, throwing an error if some expression is in fact mal-formed.

That seems to me to better fit "validatiaon" than "inspection" (the latter would imply tracing to me)

Proofs/Experiments/Abs.lean
alexkeizer: I thought the problem with `ret` is that makes it hard to cleanly define the exit state, since the PC becomes symbolic after execit? Then again, we don't use VCG here, so I guess that won't be a problem in this specific instance

Tactics/Attr.lean
shigoel: Indeed. I haven't reached `AxEffects` yet, so found the comment confusing. It'd help to add what you said to the description up above.

Arm/State.lean
alexkeizer: I wanted to put the `h` hyp in front, to avoid accidentally capturing anything, and having implicit args after an explicit one felt a bit off.
Also, there are no other arguments from which `fld`, `val`, `n\1` or `addr\1` could be inferred, so having them explicit seems the right choice

Arm/State.lean
alexkeizer: I didn't know about that file. Let's keep these lemmas here for now, and mention in that issue that we should get rid of the hlper lemmas in `State.lean` once `StateEq.lean` builds again

Arm/BitVec.lean
alexkeizer: It's not necessary, per-se, but it seems easier to have it for manual proofs (so that just `simp` will already pick up on it). It was also a marker that this might make for a good upstream simp-lemma

Arm/State.lean
alexkeizer: I'm not sure these would make good simp-lemmas: they don't really have an obvious direction that makes for a good simp-normal form

Tactics/Attr.lean
alexkeizer: I've updated the description, it now reads:
```
  -- enable extra checks for debuggins `sym_n`,
  -- see `AxEffects.validate` for more detail on what is being type-checked
  registerOption `Tactic.sym.debug {
    defValue := true
    descr := "enable/disable type-checking of internal state during execution \
      of the `sym_n` tactic, throwing an error if mal-formed expressions were \
      created"
  }

```

Arm/Insts/Common.lean
alexkeizer: Done

Arm/State.lean
alexkeizer: Done

Arm/State.lean
alexkeizer: In particular, if we had both `∀ n addr, m₁.read_bytes n addr = m₂.read_bytes n addr` and the symmetric version `∀ n addr, m₂.read_bytes n addr = m₁.read_bytes n addr` hypotheses, and include both in the simp-set, then these equalities would loop. I think nowadays simp is smart enough to detect such loops, but it's still considered bad form to have a looping simp-set

Proofs/Experiments/Abs.lean
alexkeizer: It was in fact easy to run the whole program, I've updated the proof

Arm/State.lean
shigoel: Thanks! That helps!

Arm/State.lean
shigoel: You're absolutely right, thanks!

alexkeizer:
@shigoel I think I've addressed all your comments, another round of review please!

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,590 @@
...
+
+/-- An initial `AxEffects` state which has no writes.
+That is, the given `state` is assigned to be both the initial and the current
+state -/
+def initial (state : Expr) : AxEffects where
+  initialState      := state
+  currentState      := state
+  fields            := .empty
+  nonEffectProof    :=
+    mkLambda `f .default (mkConst ``StateField) <|
```
shigoel, line 90: Could we have a Lean-equivalent comment for all the `Expr`s we create? I do wish we had a nice quoting mechanism.

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,590 @@
...
+      let pre ← fvars.mapM fun expr => do
+        let ty := lctx.get! expr.fvarId! |>.type
+        mkDecideProof ty
+
+      let nonEffectProof := mkAppN nonEffectProof pre
+      trace[Tactic.sym] "constructed: {nonEffectProof}"
+      return nonEffectProof
+
+/-- Get the value for a field, if one is stored in `eff.fields`,
+or assemble an instantiation of the memory non-effects proof -/
```
shigoel, line 180:
```suggestion
or assemble an instantiation of the non-effects proof -/
```

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,590 @@
...
+
+/-! ## Validation -/
+
+/-- type check all expressions stored in `eff`,
+throwing an error if one is not type-correct.
+
+NOTE: does not necessarily validate *which* type an expression has,
+validation will still pass if types are different to those we claim in the
+docstrings -/
+def validate (eff : AxEffects) : MetaM Unit := do
```
shigoel, line 503: A comment on why `validate` is run only under `sym.debug` would be nice. Performance reasons?

Also, another comment re. why `validate` is necessary in the first place. To debug the `Expr`s we create? Would the need for this function go away if we had a quoting mechanism?

Tactics/Sym.lean
```diff
@@ -179,6 +179,87 @@ def withoutHyp (hyp : Name) (k : TacticM Unit) : TacticM (Option FVarId) :=
...
+    eff ← eff.withField hErr.toExpr
+
+    if let some h_sp := c.h_sp? then
+      let hSp ← SymContext.findFromUserName h_sp
+      -- let effWithSp?
+      eff ← match ← eff.withStackAlignment? hSp.toExpr with
+        | some newEff => pure newEff
+        | none => do
+            trace[Tactic.sym] "failed to show stack alignment"
+            -- FIXME: in future, we'd like to detect when the `sp_aligned`
```
shigoel, line 207: Thanks for this note!

shigoel:
@alexkeizer Things look great, modulo some nits. I'm eager to merge this (esp. having having spent such a lot of time reading this!), so let's wrap this up.

Tactics/Reflect/AxEffects.lean
alexkeizer: In a bug-free world, we only ever construct well-formed expressions, and running `validate` is a waste of time. Alas, I did write some bugs, so I wrote `validate` to help me catch exactly when I construct a bad expressions. 

So, a user would essentially never want to turn on `debug`, it's only there for us, the implementors of `sym_n`, to help in debugging

Tactics/Reflect/AxEffects.lean
```diff
@@ -0,0 +1,590 @@
...
+  /-- A proof that `r <field> <currentState> = <value>` -/
+  proof : Expr
+  deriving Repr
+open AxEffects.FieldEffect
+
+/-- `AxEffects` is an axiomatic representation of effects,
+i.e., `AxEffects` transforms a description of an `ArmState` written as
+a sequence of `w` and `write_mem`s to some initial state
+into a set of hypotheses that relates reading fields from the final state
+to the initial state.
```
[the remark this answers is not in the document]
alexkeizer: I've added the following comment both here and at `fromExpr`:
```lean
Note that as soon as an unsupported expression (e.g., an `if`) is encountered,
the whole expression is taken to be the initial state,
even if there might be more `w`/`write_mem`s in sub-expressions.
```

However, `sym_n`'s implementation does additionally check that the initial state is what we expect, so added a comment there as well:
```
/- Assert that the initial state of the obtained `AxEffects` is equal to
    the state tracked by `c`.
    This will catch and throw an error if the semantics of the current
    instruction still contains unsupported constructs (e.g., an `if`) -/
```

Tactics/Reflect/AxEffects.lean
alexkeizer: Done

shigoel (approved):
Thanks, @alexkeizer!

PR #124, opened 2024-08-27
feat: Do not CSE Nat literals

bollu:
This fixes the performance scaling problems reported by @shigoel , where the tactic was performing too much work.

PR #130, opened 2024-08-28
feat: enable fuel for CSE

bollu:
In doing so, we also cleanup the logging code, since the process of adding fuel involves visiting each call site, which is exactly where we want to add logging information! 

Supercedes #113, as the fixes from that PR have been merged in. Foo #124 , #114


Proofs/SHA512/SHA512Sym.lean
```diff
@@ -47,6 +47,7 @@ abbrev ktbl_addr : BitVec 64 := 0x1b4300#64
 -- set_option profiler true in
 -- set_option profiler.threshold 1 in
 -- set_option pp.deepTerms false in
+set_option trace.Tactic.cse.summary true in
 theorem sha512_block_armv8_1block (s0 sf : ArmState)
   -- (FIXME) Ignore the `stp` instruction for now.
   (h_s0_pc : read_pc s0 = 0x1264c4#64)
```
alexkeizer, line 53: I'm not a big fan of having trace options committed, since they make tend to make the build log quite noisy

Tactics/CSE.lean
```diff
@@ -104,6 +104,13 @@ structure CSEConfig where
   minOccsToCSE : Nat := 2
   /-- Whether the tactic should throw an error if no CSEable subterms were found. -/
   failIfUnchanged : Bool := true
+  /-- Number of steps the tactic should spend searching for subterms to gather information. -/
+  fuelSearch : Nat := 1000
+  /-- Number of steps the tactic should spend performing subexpression elimination. -/
+  fuelEliminate : Nat := 1000
+  /-- Whether we are performing a dry run, where we will identify repeated
+      subterms, and simulate the sharing algorithm, but perform no actual sharing. -/
+  dryRun? : Bool := false
```
alexkeizer, line 113: Minor not: I tend to use `?` as a suffix to indicate something is an `Option`, so I'd prefer calling this just `dryRun`, or `isDryRun` if you want to emphasize it's a Bool

Tactics/CSE.lean
```diff
@@ -125,13 +132,10 @@ structure State where
   a counter to generate new names.
   -/
   gensymCount : Nat := 1
-
-instance : ToMessageData State where
-  toMessageData s := Id.run do
-    let mut msg := m!""
-    for (k, v) in s.canon2data.toArray do
-      msg := msg ++ Format.line ++ m!"{k} := {repr v}"
-    msg
```
alexkeizer, line 134: Why did this instance disappear?

Tactics/CSE.lean
```diff
@@ -104,6 +104,13 @@ structure CSEConfig where
   minOccsToCSE : Nat := 2
   /-- Whether the tactic should throw an error if no CSEable subterms were found. -/
   failIfUnchanged : Bool := true
+  /-- Number of steps the tactic should spend searching for subterms to gather information. -/
+  fuelSearch : Nat := 1000
+  /-- Number of steps the tactic should spend performing subexpression elimination. -/
+  fuelEliminate : Nat := 1000
```
alexkeizer, line 110: Minor nit: I think `searchFuel` and `eliminateFuel` would read more naturally.
to me `hasFuelSearch` communicates searching for fuel, while
`hasSearchFuel` communicates having fuel for searching (which is the intended meaning).

Very subjective, though, so feel free to ignore

Tactics/CSE.lean
```diff
@@ -144,8 +148,43 @@ def getState : CSEM State := get
...
+
+def CSEM.consumeFuelSearch : CSEM Unit :=
+  modify fun s => { s with currentFuelSearch := s.currentFuelSearch - 1 }
+
+def CSEM.hasFuelEliminate? : CSEM Bool := do
+  return (← getState).currentFuelEliminate > 0
+
+def CSEM.consumeFuelEliminate : CSEM Unit :=
+  modify fun s => { s with currentFuelEliminate := s.currentFuelEliminate - 1 }
+
```
alexkeizer, line 170:
```suggestion
```
extraneaous whitespace

Tactics/CSE.lean
```diff
@@ -144,8 +148,43 @@ def getState : CSEM State := get
...
+    (collapsed : Bool := true)
+    (traceClass : Name := `Tactic.cse.summary) : CSEM α :=
+  Lean.withTraceNode traceClass (fun _ => return header) k (collapsed := collapsed)
+
+/-- Create a trace note that folds `header` with `(NOTE: can be large)`,
+and prints `msg` under such a trace node.
+-/
+def CSEM.traceLargeMsg (header : MessageData) (msg : MessageData) : CSEM Unit :=
+    withTraceNode m!"{header} (NOTE: can be large)" do
+      trace[Tactic.cse.summary] msg
```
alexkeizer, line 187: Nice pattern! I should steal these helpers for `sym`

Tactics/CSE.lean
```diff
@@ -144,8 +148,43 @@ def getState : CSEM State := get
...
+  currentFuelSearch := cfg.fuelSearch,
+  currentFuelEliminate := cfg.fuelEliminate,
+}
+
 def CSEM.run (val : CSEM α) (config : CSEConfig) : TacticM α :=
-   val.run' {} |>.run config
+   val.run' (State.ofConfig config) |>.run config
+
+def CSEM.hasFuelSearch? : CSEM Bool := do
+  return (← getState).currentFuelSearch > 0
```
alexkeizer, line 160:
```suggestion
def CSEM.hasFuelSearch : CSEM Bool := do
  return (← getState).currentFuelSearch > 0
```

For the same reason as `dryRun`: the `has` prefix already communicates the Booleannes

Tactics/CSE.lean
```diff
@@ -144,8 +148,43 @@ def getState : CSEM State := get
...
+
+def CSEM.hasFuelEliminate? : CSEM Bool := do
+  return (← getState).currentFuelEliminate > 0
+
+def CSEM.consumeFuelEliminate : CSEM Unit :=
+  modify fun s => { s with currentFuelEliminate := s.currentFuelEliminate - 1 }
+
+
+def CSEM.isDryRun? : CSEM Bool := do
+  return (← getConfig).dryRun?
```
alexkeizer, line 173:
```suggestion
def CSEM.isDryRun : CSEM Bool := do
  return (← getConfig).dryRun?
```

Tactics/CSE.lean
```diff
@@ -144,8 +148,43 @@ def getState : CSEM State := get
...
+   val.run' (State.ofConfig config) |>.run config
+
+def CSEM.hasFuelSearch? : CSEM Bool := do
+  return (← getState).currentFuelSearch > 0
+
+def CSEM.consumeFuelSearch : CSEM Unit :=
+  modify fun s => { s with currentFuelSearch := s.currentFuelSearch - 1 }
+
+def CSEM.hasFuelEliminate? : CSEM Bool := do
+  return (← getState).currentFuelEliminate > 0
```
alexkeizer, line 166:
```suggestion
def CSEM.hasFuelEliminate : CSEM Bool := do
  return (← getState).currentFuelEliminate > 0
```

Tests/Tactics/CSE.lean
```diff
@@ -15,38 +15,147 @@ We want to check that it correctly ees that there are:
...
-    y := { occs := 16, size := 1 }
-    y + y + (y + y) + (y + y + (y + y)) + (y + y + (y + y)) := { occs := 1, size := 23 }
-    y + y + (y + y) + (y + y + (y + y)) := { occs := 1, size := 15 }
-    y + y + (y + y) := { occs := 4, size := 7 }
-    x + x := { occs := 1, size := 3 }
-    x := { occs := 2, size := 1 }
-    x + x + (y + y + (y + y)) := { occs := 1, size := 11 }
-    y + y := { occs := 8, size := 3 }
-[Tactic.cse.generalize] ⏭️ Skipping y + y + (y + y) + (y + y + (y + y)) +
-      (y + y + (y + y)): Unprofitable { occs := 1, size := 23 } .
```
alexkeizer, line 33: I'm not sure how valuable having tests with tracing enabled is: I'm certainly not confident that I can review this diff and say for sure that the new trace does what it's supposed to. I imagine we might want to tweak the tracing pretty often, and if we just blindly accept the new trace as the expected output every time, the test won't catch any "bugs" in the new trace anyway

Tactics/CSE.lean
```diff
@@ -228,58 +272,81 @@ partial def CSEM.planCSE (e : Expr): CSEM GeneralizeArg := do
 Plan to perform a CSE for this expression, by building a 'GeneralizeArg'.
 -/
 def CSEM.generalize (arg : GeneralizeArg) : CSEM Bool := do
+  consumeFuelEliminate
+  unless (← hasFuelEliminate?) do
+    trace[Tactics.cse.summary] "⏸️ CSE ran out of fuel while performing subexpression elimination. increase `fuelEliminate`."
+    return false
```
alexkeizer, line 278: It seems we pretty much only call `consumeFuel` and `hasFuel` together. We could slightly simplify these callsites by having a single function that consumes a fuel, and returns whether there is any left at the same time, at the cost of this function being slightly less self-explanatory than the current split in two functions.

Just a thought, not sure it matters much either way

alexkeizer (commented):
I like the extra tracing! I've mostly got some comments about names.

Otherwise, the diff in the test is pretty unreviewable, and I'm not convinced having traces in the test is worth the hassle, but there's nothing I'd consider majorly blocking here

Tactics/CSE.lean
```diff
@@ -104,6 +104,13 @@ structure CSEConfig where
   minOccsToCSE : Nat := 2
   /-- Whether the tactic should throw an error if no CSEable subterms were found. -/
   failIfUnchanged : Bool := true
+  /-- Number of steps the tactic should spend searching for subterms to gather information. -/
```
pennyannn, line 107: What is the motivation for having two separate fuels `searchFuel` and `eliminateFuel`? Could they be represented as one fuel?

Tactics/CSE.lean
alexkeizer: Yeah, I had trouble coming up with a sufficiently clean yet self-explanatory name for the combined method, which is a good hint the functionality might be too magical. Let's leave it as-is then, the duplication is not too bad

pennyannn (commented):
One high-level comment, several files I found are without headers. Could you add them?

Tactics/CSE.lean
bollu: @alexkeizer because I stopped using the instance, it was far too verbose; I opt for more controlled printing now.

Tests/Tactics/CSE.lean
bollu: @alexkeizer I agree, I see the current situation as a 'stop-gap', until we have the ability to look for patterns in guard messages. At least for me, it's easy to skim and see that the CSE is 'doing the right thing as per vibes', so I'd opt to keep it in.

Tactics/CSE.lean
bollu: Changed accordingly.

Tactics/CSE.lean
bollu: Changed accordingly.

Tactics/CSE.lean
bollu: I added some text for why it's useful to have separate fuels. Conceptually, the algorithm has two phases, and it's useful for both debugging and for controlling behaviour to be able to fuel these separately.

Proofs/SHA512/SHA512Sym.lean
bollu: Discussed alternatively.

bollu:
@shigoel ready for merge!

Proofs/SHA512/SHA512Sym.lean
```diff
@@ -47,6 +47,7 @@ abbrev ktbl_addr : BitVec 64 := 0x1b4300#64
 -- set_option profiler true in
 -- set_option profiler.threshold 1 in
 -- set_option pp.deepTerms false in
+set_option trace.Tactic.cse.summary true in
```
shigoel, line 50: I'm wary of this -- there's a lot of `make` output already from `simp_mem`. Could you tell me what the output here looks like?

shigoel: Also, if there's no `#guard_msgs` here, I'd rather this was turned off.

Tactics/CSE.lean
```diff
@@ -104,6 +104,16 @@ structure CSEConfig where
...
+  fuelSearch : Nat := 1000
+  /-- Number of steps the tactic should spend performing subexpression elimination.
+   It can be useful to have large amounts of fuel for searching, and very little for eliminating,
+   to search for maximal subterm sharing, and to then eliminiate the most common occurrences.
+  -/
+  fuelEliminate : Nat := 1000
+  /-- Whether we are performing a dry run, where we will identify repeated
+      subterms, and simulate the sharing algorithm, but perform no actual sharing. -/
+  dryRun? : Bool := false
```
shigoel, line 116: Add user-guidance for when to use `dryRun?`, and how it reports the info. it gleans.

Tests/Tactics/CSE.lean
```diff
@@ -1,52 +1,165 @@
...
-We want to check that it correctly ees that there are:
+We want to check that it correctly sees that there are:
 - `16` occurrences of `y`,
 - `8` occurrences of `y + y`,
 - `4` occurrences of `(y + y) + (y + y)`.
 
 -/
 
-set_option trace.Tactic.cse.generalize true in
 set_option trace.Tactic.cse.summary true in
```
shigoel, line 22: Does setting this option give a verbose `make` output?

shigoel:
@bollu Looks good in general, but I'm concerned whether the traces give a verbose `make` output.

Tactics/CSE.lean
```diff
@@ -169,41 +210,46 @@ The function is partial because of the call to `tryAddExpr` that
...
-    return .some data
-  | .none =>
-    let data := { occs := 1, size : ExprData }
-    setState {
-      s with
-      canon2data := s.canon2data.insert e data,
-    }
-    trace[Tactic.cse.collection] m!"added new '{e}' with info '{repr data}'" -- TODO: make this a trace node child.
-    return .some data
+  withTraceNode m!"({e}):({t}) [relevant? {if relevant? then checkEmoji else crossEmoji}] (unfold for subexpressions...)" do
```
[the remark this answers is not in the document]
bollu: Aha, it's a `traceNode`, so it's saying, "if you really want to see what subexpressions I generated, please click me to unfold

Proofs/SHA512/SHA512Sym.lean
bollu: Agreed.

bollu: I removed the trace.

Tactics/CSE.lean
bollu: @shigoel I wrote that this meant for debugging. I'd rather not write down how it reports the info, since I could see that evolving with the tactic --- the implementation essentially writes down what it skipped due to it being a dry run.

Tests/Tactics/CSE.lean
bollu: No, because it's guarded by a `#guard_msgs`.

bollu:
@shigoel I understand, I cleaned up the one that wasn't guarded. All of the other trace messages are `#guard_msgs`, so we should be good!

bollu:
@shigoel ready for merge, methinks!

Tactics/CSE.lean
shigoel: Alright -- that works.

shigoel:
Thanks, @bollu!

shigoel (approved):

PR #133, opened 2024-08-29
feat: simp-based aggregation of effects

alexkeizer:
This PR is stacked upon #129 

### Description:

Change `AxEffect` and `SymContext` to track all axiomatic effect hypothesis that are added during symbolic evaluation, and feed these to a `simp` post-processing step to implement a basic form of effect aggregation.
This replaces the manual `simp`ing we've been doing so far, and works well enough for the examples we've been interested in so far, but we might want to consider a more principled approach to effect aggregation when we run into scalability issues. 

Additionally, to make the aggregation work properly, we add a pass that checks whether symbolic evaluation has reached the final state, and does the appropriate substitution (replacing `sf` with the last intermediate state we introduced).
* For now, this check is relatively simple, and only works when the amount of steps to run was a concrete literal, that reached 0 after simulation. We could integrate this with the `whileTac` approach and accept a user-extensible tactic to prove that `runSteps = 0`.
* We add initial `h_program`, `h_pc` and `h_err` hypotheses to the post-processing simp-set, so that goals like "the step after simulating for n steps is error-free" will be closed entirely automatically by calling `sym_n $n`
* We also expose a separate `sym_aggregate` tactic, which will do a type-based search of the local context to determine (non-)effect hypotheses. This works better when simulation is spread out over multiple `sym_n` calls.
* Additionally, now that we have a good aggregation story, we no longer expose `h_step` lemmas in the goal states, making those a tad less cluttered

### Testing:

 `make all` ran locally

### License:


alexkeizer:
@shigoel I merged main, and fixed the breaks that occured

shigoel:
High-level comments (caveat: before I've read the code, so I may come back and modify this if it's clear from the implementation):

> Additionally, to make the aggregation work properly, we add a pass that checks whether symbolic evaluation has reached the final state, and does the appropriate substitution (replacing sf with the last intermediate state we introduced).

As we discussed, it'd be helpful to have a way of saying "aggregate _now_" as opposed to aggregate at the final state. Is that doable right now, or in a future PR?

> We also add initial h_program, h_pc and h_err hypotheses to the post-processing simp-set, so that goals like "the step after simulating for n steps is error-free" will be closed entirely automatically by calling sym_n $n

This is pretty cool!

Proofs/Experiments/Abs/AbsVCGTandem.lean
shigoel, line 1: Thanks for fixing these proofs! I need to rewrite this using the new VCG stuff, so it'll get cleaner!

Proofs/Popcount32.lean
```diff
@@ -79,18 +79,6 @@ theorem popcount32_sym_no_error (s0 s_final : ArmState)
   simp_all only [state_simp_rules, -h_run]
   -- Symbolic Simulation
   sym_n 27
```
shigoel, line 81: Correct me if I'm wrong, but `sym_n` still leaves `h_step_<i>` hyps. in the context, right? If so, are we ready to reduce goal clutter by clearing them away? After all, we have effects and non-effects hyps now.

shigoel: Awkward comment placement because the diff. reminded me of this...

Tactics/Sym.lean
```diff
@@ -409,3 +410,39 @@ Did you remember to generate step theorems with:
...
+            {sfEq} in:\n{goal}"
+
+        let goal ← subst goal hEqId
+        trace[Tactic.sym] "performed subsitutition in:\n{goal}"
+
+        replaceMainGoal [goal]
+
+    -- Rudimentary aggregation: we feed all the axiomatic effect hypotheses
+    -- added while symbolically evaluating to `simp`
+    withMainContext <| do
```
shigoel, line 439: I see that aggregation indeed happens only when we reach the final state because the theorem's conclusion is typically in its terms (anyway the user doesn't have a way to refer to the intermediate states when stating the theorem). `axHyps` won't help in simplifying the conclusion unless the effects of the last instruction are also in the context.

Proofs/Experiments/Abs/AbsVCGTandem.lean
alexkeizer: I wonder... I noticed while fixing the AbsVCGTandem proof that the aggregation does not really work there, because we simulate one step at a time. The current final simp-postprocessing of the goal works best if we only have a single `sym_n` call, because we don't remember the `axHyps` of earlier calls

Proofs/Popcount32.lean
alexkeizer: We could, except for the fact that this aggregation does not quite work cleanly with the VCG framework, where we simulate one step at a time. We could manually invoke the right (non-)effects hyps, but that is somewhat annoying to do by hand

Proofs/Experiments/Abs/AbsVCGTandem.lean
shigoel: We do simulate one step at a time for those proofs, yes. What would help there is what we already plan to do: have a way of aggregating at any given point in the simulation, not only when we've reached the end. 

Another thing that is worth considering is having pre- and post-processing hooks in `sym_n` so that we can insert our own tactic blocks before and after the simulation of each instruction (e.g., tactics preceding the invocation of `sym_n 1 at ...` in the AbsVCGTandem proof). Can the `while` bit of `sym_n` can help with some of that already?

Also, @bollu -- thoughts here, since you're a VCG user too?

shigoel: To be clear, the scope of this discussion goes beyond this PR (which I'm happy to accept in a bit), but the PR provides a good jumping off point for this discussion.

Proofs/Popcount32.lean
shigoel: `sym_n` does generate `h_s{i}_<component>` hyps. that describe the `h_step_{i}` hyps in terms of reads, right? Those `h_s{i}_<component>` ought to work for manual aggregation needed in VCG proofs at the moment. 

The AbsVCGTandem proof would need to change to stop using `h_step_{i}` hyps. for aggregation and use `h_s{i}_<component>` hyps instead. I can try that after this PR's merged, and if I find myself not needing the `h_step_{i}` hyps at all, then let's clear them out.

alexkeizer:
@shigoel based on your comments, I've also implemented a stand-alone aggregation tactic `sym_aggregate`, which will look up all non-effects theorems in the local context (based on types), and use those for a simplification step. This works better for proofs where we have seperate `sym_n` steps.

With this, we should be able to get rid of `h_step_` lemmas, I'll have a look if that's easy to do, or best left for a follow-up

Proofs/Experiments/Abs/AbsVCGTandem.lean
alexkeizer: This has now been fixed, by adding a separate `sym_aggregate` tactic

Tactics/Sym.lean
alexkeizer: We could do some rudimentary intermediate aggregation, by doing a `simp_all` instead of only simplifying the goal. That feels like it could be much more brittle, though, which is why I shied away from that approach

shigoel:
@alexkeizer Thanks for clearing away the `h_step`s! So very nice!

`sym_aggregate` still aggregates the effects when the final state is reached, is that right? I meant aggregation of the effects in the middle of a long simulation (e.g., suppose we need to simulate 200 instructions and we'd like to aggregate the effects every 20 instructions -- we can invoke something like `sym_n 20 at s0; sym_aggregate at s20`). Would `simp_all` really be needed then? Could `simp` with `axHyps` at `<h_s20>` or some such not work?

Also, discussion for the future: 

> Another thing that is worth considering is having pre- and post-processing hooks in sym_n so that we can insert our own tactic blocks before and after the simulation of each instruction (e.g., tactics preceding the invocation of sym_n 1 at ... in the AbsVCGTandem proof). Can the while bit of sym_n can help with some of that already?

shigoel:
I am going to hit merge so that we can start using this new feature, but I'll spawn off an issue to keep track of our conversations re. `sym_n`'s evolution. Thanks, @alexkeizer!

PR #136, opened 2024-09-03
feat: prove specification of Max program that manipulates memory via new VCG

bollu:
Supercedes #128, and have broken down the patches into a patch sequence that we can peel commits from.


Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Shilpi Goel
+
+Experimental: Use the Correctness module to prove that this program computes the
```
pennyannn, line 6: The comment seems to be referring to another program?

Arm/Insts/Common.lean
```diff
@@ -55,6 +55,32 @@ def AddWithCarry (x : BitVec n) (y : BitVec n) (carry_in : BitVec 1) :
   let V := if signExtend (n + 1) result = signed_sum then 0#1 else 1#1
   (result, (make_pstate N Z C V))
 
+/-- When the carry bit is `0`, `AddWithCarry x y 0 = x + y` -/
```
pennyannn, line 58: The number of lemmas seem to grow in this file. Should we break it into Sections? Better ideas?

Arm/State.lean
```diff
@@ -290,6 +290,45 @@ def r (fld : StateField) (s : ArmState) : (state_value fld) :=
...
+@[state_simp_rules]
+def ArmState.C (s : ArmState) : BitVec 1 := r (StateField.FLAG PFlag.C) s
+
+@[state_simp_rules]
+def ArmState.Z (s : ArmState) : BitVec 1 := r (StateField.FLAG PFlag.Z) s
+
+@[state_simp_rules]
+def ArmState.N (s : ArmState) : BitVec 1 := r (StateField.FLAG PFlag.N) s
+
+def ArmState.r_GPR_0_eq_x0 (s : ArmState) : r (StateField.GPR 0) s = s.x0 := by rfl
```
pennyannn, line 317: It's not immediate clear what these definitions are good for. Maybe consider adding a comment.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
...
+
+namespace MaxTandem
+open Max
+
+
+section PC
+
+/--# We define scoped notation for our cutpoint PCs to use in pattern matching. -/
+
+scoped notation "entry_start" => 0x894#64
```
pennyannn, line 27: Cool.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
...
+scoped notation "then_start" => 0x8b0#64
+scoped notation "else_start" => 0x8bc#64
+scoped notation "merge_end" => 0x8cc#64
+
+
+namespace ArmStateNotation
+
+/-! We build a notation for `read_mem_bytes $n $base $s` as `$s[$base, $n]` -/
+@[inherit_doc read_mem_bytes]
+syntax:max term noWs "[" withoutPosition(term)  ","  withoutPosition(term) noWs "]" : term
```
pennyannn, line 37: Cool

Arm/State.lean
pennyannn: These dot notations are convenient, but there are more than 4 gprs and also sfps. I'm wondering if there is a better way to do this than listing them out.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
...
+def entry_start_inv (s0 si : ArmState) : Prop := si = s0
+
+/-- Invariant right at the conditional branch -/
+def entry_end_inv (s0 si : ArmState): Prop :=
+  read_err si = .None ∧
+  si.program = program ∧
+  -- (FIXME) We don't really need the stack pointer to be aligned, but the
+  -- `sym_n` tactic expects this. Can we make this optional?
+  CheckSPAlignment si ∧
+  si[s0.sp - 32#64 + 12#64, 4] = s0.x0.zeroExtend 32 ∧
```
pennyannn, line 91: I like this notation. I wonder if this feature could be made global at some point.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
...
+  si[s0.sp - 32#64 + 8#64, 4] = s0.x1.zeroExtend 32 ∧
```
pennyannn, line 92: There is a function called `read_gpr` that takes a `width` argument. That function will do `zeroExtend` to the `width`. I recommend using it to hide the `zeroExtend`s so that it is easier to read.

Arm/Insts/Common.lean
bollu: @pennyannn I think we should add sections, indeed, but I'd prefer to do this in a subsequent PR where I add all the sections in one go! I made an issue to track this: https://github.com/leanprover/LNSym/issues/139

Arm/State.lean
bollu: @pennyannn I don't immediately know of a better way that is as simple as rolling our sleeves and listing them out :) Metaprogramming games can be played, but my feeling is that the complexity isn't worth it.

Proofs/Experiments/Max/MaxTandem.lean
bollu: Yes, once we gain some more experience with notation, we can move it out into a new file. I track this in an issue: https://github.com/leanprover/LNSym/issues/140

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,899 @@
...
+  · decide
+
+/--
+info: 'MaxTandem.program.stepi_0x8c8_cut' depends on axioms: [propext, Classical.choice, Lean.ofReduceBool, Quot.sound]
+-/
+#guard_msgs in #print axioms program.stepi_0x8c8_cut
+
+end CutTheorems
+
+/--
```
pennyannn, line 578: I wonder if these tactics should be moved somewhere else, like in Tactics.

Proofs/Experiments/Max/MaxTandem.lean
bollu: @pennyannn on the contrary, I prefer this, because it makes it clear that we're reading from `x0` /`x1`. I'm happy to have an alternative that name that hides the zero extend, but I'd like to keep the `s0.x0` / `s0.x1` for the massive boost in readability :)

Proofs/Experiments/Max/MaxTandem.lean
bollu: I wasn't sure, because these felt specialized to this particular proof style, which will change when @alexkeizer merges in the shiny new symbolic simulation :) I'd be happy to move them to a separate file when we use these tactics multiple times. Does that sound fair to you?

Arm/State.lean
pennyannn: Understandable.

Proofs/Experiments/Max/MaxTandem.lean
pennyannn: Could we invent new notations like `s0.w0` (Arm notation) which will stand for `s0.x1.zeroExtend 32`?

pennyannn: At first I find the `zeroExtend`s a bit distracting when reading the proof, and I knew we have an abstraction `read_gpr` for hiding it. However, as I read more into the bulk of correctness proof, I found that you have lots of reference to `zeroExtend` in later proofs as well. So now I'm not sure if you needed to expand and expose `zeroExtend` anyways. I will let you decide if more abstraction could be done then...

Arm/State.lean
```diff
@@ -702,6 +741,18 @@ def Memory.read (addr : BitVec 64) (m : Memory) : BitVec 8 :=
 
 theorem ArmState.read_mem_eq_mem_read : read_mem addr s = s.mem.read addr := rfl
 
+/-- `w` does not effect memory. -/
```
alexkeizer, line 744:
```suggestion
/-- `w` does not affect memory. -/
```

Arm/State.lean
```diff
@@ -702,6 +741,18 @@ def Memory.read (addr : BitVec 64) (m : Memory) : BitVec 8 :=
...
+@[memory_rules, state_simp_rules]
+theorem ArmState.mem_w_eq_mem (fld : StateField) (v : state_value fld) (s : ArmState) :
+    (w fld v s).mem = s.mem := by
+  cases fld <;>
+  unfold w write_base_error
+    write_base_gpr
+    write_base_sfp
+    write_base_pc
+    write_base_flag <;>
+    simp
```
alexkeizer, line 754:
```suggestion
  cases fld <;> (
    unfold w write_base_error
      write_base_gpr
      write_base_sfp
      write_base_pc
      write_base_flag
    simp)
```

Also, couldn't we have `simp` do the unfolding here as well?

Arm/State.lean
alexkeizer: These should probably be `abbrev`s

Arm/Insts/Common.lean
```diff
@@ -93,6 +156,13 @@ def Aligned (x : BitVec n) (a : Nat) : Prop :=
 instance : Decidable (Aligned x a) := by
   cases a <;> simp [Aligned] <;> infer_instance
 
+theorem Aligned_BitVecSub_64_4 {x : BitVec 64} {y : BitVec 64}
```
shigoel, line 159: Let's add the `state_simp_rules` tag to the three `Aligned`-related theorems.

Arm/Insts/Common.lean
```diff
@@ -55,6 +55,32 @@ def AddWithCarry (x : BitVec n) (y : BitVec n) (carry_in : BitVec 1) :
   let V := if signExtend (n + 1) result = signed_sum then 0#1 else 1#1
   (result, (make_pstate N Z C V))
 
+/-- When the carry bit is `0`, `AddWithCarry x y 0 = x + y` -/
+theorem fst_AddWithCarry_eq_add (x : BitVec n) (y : BitVec n) :
```
shigoel, line 59: Let's also add the `state_simp_rules` tag to these `AddWithCarry` theorems.

Though... I'm getting worried about bloat there. We should consider defining another `simp` attr that can help in simplifying a goal state after symbolic simulation is done.

Proofs/Experiments/Max/MaxTandem.lean
shigoel: If we are building notations, then I'd prefer defining `s0.w0` and the like to stay true to Arm.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -0,0 +1,898 @@
...
+      for name in names.getElems do
+        if name.getId.toString.isPrefixOf hyp.userName.toString then
+          replaceMainGoal [← (← getMainGoal).clear hyp.fvarId]
+          continue
+
+
+
+section ForLean
+
+theorem BitVec.not_slt {w} (a b : BitVec w) : ¬ (a.slt b) ↔ (b.sle a) := by
```
shigoel, line 613: More to Arm/Bitvec.lean with a TODO to upstream?

shigoel:
Looks good in general, and I'd be interested to see whether you can reuse the symbolic simulation done during the partial correctness proof to also prove termination. See `Correctness.by_the_method` and the `AddLoopTandem` proof.

bollu:
> and I'd be interested to see whether you can reuse the symbolic simulation done during the partial correctness

I too would be interested, but I would prefer to do this after I prove `memcpy` partially correct :) That feels slightly higher up on the priority queue for me, since it's a target program that we promised to verify!

Arm/Insts/Common.lean
bollu: I am not convinced that these are good `simp` rues, because as far as I can tell, we will almost never never naturally have the proof that `x` and `y` are aligned! This will almost always require proof search, as far as I can tell?

Arm/Insts/Common.lean
bollu: @shigoel same concern :) tagging these as `state_simp_rules` means I need to change `AddTandem`, so I'd rather do that in a separate PR. Does that work for you?

Proofs/Experiments/Max/MaxTandem.lean
bollu: Done!

bollu:
@shigoel done, I've fixed up the comments that I felt were actionable :)

Arm/Insts/Common.lean
shigoel: Aha, good point! We only use `Aligned` in `CheckSPAlignment` right now, and since we're going to have automation for the latter, I rescind my comment.

shigoel (approved):

PR #143, opened 2024-09-05
fix: better handling of unconditional branches 

alexkeizer:
### Description:

We now attempt to reflect the PC after each step, only falling back to blindly incrementing if reflection failed.
With this change, `sym_n` is able to deal with unconditional branches, and I've fixed the `AddLoop` example accordingly.

The `reflectBitVecValue` used here does involve a `whnfD`, which might be a expensive.
However, this reduction only happens when the value of the PC is not a literal (i.e., for conditional branches), where reflection of the PC really is necessary.
On non-branching instructions, or even unconditional branches, the call should be basically free, 
so there is little downside to this approach.

### Testing:

`make all` ran locally

### License:


Tactics/Sym.lean
```diff
@@ -181,7 +181,11 @@ def withoutHyp (hyp : Name) (k : TacticM Unit) : TacticM (Option FVarId) :=
 
 /-- Given an equality `h_step : s{i+1} = w ... (... (w ... s{i})...)`,
 add hypotheses that axiomatically describe the effects in terms of
-reads from `s{i+1}` -/
+reads from `s{i+1}`.
+
+Return the context for the next step (see `SymContext.next`), where
+we attempt to determine the new PC by reflecting the obtained effects,
+falling back to incrementing the PC if reflaction failed. -/
```
shigoel, line 188:
```suggestion
falling back to incrementing the PC if reflection failed. -/
```

Proofs/Experiments/AddLoop/AddLoop.lean
```diff
@@ -306,21 +306,8 @@ theorem effects_of_nextc_from_0x4005a4
...
-  r StateField.PC s1 = 0x4005b0#64 : Prop
-  but is expected to have type
-  r StateField.PC s1 = 0x4005a8#64 : Prop
-  -/
-  -- sym_n 2
-  sym_n 1 at s0
-  sym_n 1 at s1
-  simp (config := {ground := true}) only [*, state_simp_rules, bitvec_rules, minimal_theory]
+  sym_n 2
+  simp (config := {ground := true}) only []
```
shigoel, line 310: Sweet

Tactics/Sym.lean
```diff
@@ -249,7 +253,19 @@ def explodeStep (c : SymContext) (hStep : Expr) : TacticM SymContext :=
         eff.validate
 
       eff.addHypothesesToLContext s!"h_{c.next_state}_"
-    return { c with axHyps := axHyps ++ c.axHyps}
+    let c := { c with axHyps := axHyps ++ c.axHyps}
+
+    -- Attempt to reflect the new PC
+    let nextPc ← eff.getField .PC
+    let nextPc? ← try
+      let nextPc ← reflectBitVecLiteral 64 nextPc.value
```
shigoel, line 261: It'd be good to add a note here re. potential performance issues of using `reflectBitVecLiteral`, like you mentioned in the summary.

shigoel:
@alexkeizer Looks good, but I'd like to see a perf comparison to know the cost of figuring out the next PC this way. Could you run that experiment, please?

Tactics/Sym.lean
alexkeizer: Done

alexkeizer:
> @alexkeizer Looks good, but I'd like to see a perf comparison to know the cost of figuring out the next PC this way. Could you run that experiment, please?

I ran `#time` on gcm_gmult_v8, and it took basically the same time as before (450ms). This checks out: `reflectBitVecLiteral` will only do a reduction if the value is not obviously a constant, but our stepi generation will have reduced the value to a constant wherever possible (i.e., for everything besides conditional branches).

So the optimization I suggested would not make much of a difference: all the reflections we could skip by keeping that list are exactly the calls to `reflectBitVecLiteral` that are cheap, basically free, already.

shigoel:
@alexkeizer Good to know, and thanks for running the numbers.

shigoel (approved):

PR #144, opened 2024-09-06
Clean up of AddLoop and Abs VCG proofs

shigoel:
### Description:

Remove the `SP` Alignment checks from AddLoop and Abs VCG proofs. Also use `Correctness.by_the_method` to prove `Abs` correct, with a `sorry` left for @alexkeizer to look at.

### Testing:

`make all` ran locally.

### License:


Proofs/Experiments/Abs/AbsVCGTandem.lean
```diff
@@ -253,42 +276,35 @@ theorem partial_correctness :
...
-    split <;> bv_decide
+    constructor
+    · -- Partial Correctness Proof
+      split <;> bv_decide
+    · -- Termination Proof
+      apply Exists.intro 4
+      simp only [Spec'.cut, abs_cut, rank,
+                 state_simp_rules]
+      have h_run : run 4 s0 = s4 := by
+        -- (TODO @alex) If we kept track of the number of steps we took during
```
shigoel, line 301: @alexkeizer Here's a request for you!

Proofs/Experiments/Abs/AbsVCGTandem.lean
```diff
@@ -101,13 +100,16 @@ theorem program.stepi_0x4005d0_cut (s sn : ArmState)
...
+  r (StateField.GPR 1#5) sn =
+    (BitVec.zeroExtend 64
+      (BitVec.zeroExtend 32 (r (StateField.GPR 0x0#5) s))) ∧
+  (∀ (f : StateField), f ≠ StateField.PC → f ≠ (StateField.GPR 1#5) →
+                        r f sn = r f s) ∧
   r StateField.PC sn = 0x4005d4#64 ∧
   r StateField.ERR sn = .None ∧
-  sn.program = program ∧
-  CheckSPAlignment sn := by
+  sn.program = program:= by
```
alexkeizer, line 112:
```suggestion
  sn.program = program := by
```

Proofs/Experiments/Abs/AbsVCGTandem.lean
alexkeizer: We do track the steps, thats why I added the `stepi s{i} = s{i+1}` local hypotheses!

At first I thought you wanted to keep `h_run` around with its original type, but we'd have to aggregate here anyway, because we only simulate one step at a time, so it'd be effectively the same as just using the `stepi_s{j}` lemmas

shigoel: Thanks! I'd accidentally cleared those `stepi_` hyps and then forgotten all about them!

PR #148, opened 2024-09-09
feat: change `AxEffects.toSimpTheorems` to generate more specific discrtree keys

alexkeizer:
### Description:

The method we were using to construct simp-lemmas sets a `noIndexAtArgs` option, meaning that a lemma like 
`r (.GPR 0#5) s10` would get indexed *only* on `r`, and the arguments were ignored. This caused a lot of attempts at unification.

Ideally, we'd want `mkSimpTheorems` to expose `noIndexAtArgs` as an option, so we could disable it. 
For now, we workaround the issue with a postprocessing step that regenerates the discrtree keys with this behaviour turned off. Sadly, the failed unifications turn out to not be the main expense: with this change we can simulate SHA512 for 190 steps, which is a relatively small improvement over before.

Then, I set the priority of the field-specfic lemmas higher than the default priority used for the nonEffectProof, to prevent simp from trying to discharge the expensive side-conditions when we already know a cheaper lemma that is applicable. This did give an impressive speedup.

All together, we are now able to simulate 250 steps of SHA512 with the default heartbeat limit (when tracing is turned off! Tracing does consume some heartbeats, so with tracing the limit is lower)

### Testing:

`make all` succeeded locally

### License:


alexkeizer:
By the way, let's not delete this branch. I'm writing a document with some benchmarks I've run, noting down the exact commit IDs I used for the benchmarks, but those are commits which haven't landed in main yet

Tactics/Reflect/AxEffects.lean
```diff
@@ -661,19 +662,53 @@ def toSimpTheorems (eff : AxEffects) : MetaM SimpTheorems := do
...
+        )
+
+        pure <| newThms.foldl addSimpTheoremEntry thms
 
     let mut thms : SimpTheorems := {}
 
     for ⟨field, {proof, ..}⟩ in eff.fields do
-      thms ← add thms proof s!"field_{field}"
+      /- We give the field-specific lemmas a high priority, since their
+      applicacibility is determined entirely by discrtree matching.
```
shigoel, line 706:
```suggestion
      applicability is determined entirely by discrtree matching.
```

Arm/MinTheory.lean
```diff
@@ -8,10 +8,10 @@ attribute [minimal_theory] ite_false
 attribute [minimal_theory] dite_true
 attribute [minimal_theory] dite_false
 attribute [minimal_theory] ite_self
-attribute [minimal_theory] and_true
-attribute [minimal_theory] true_and
-attribute [minimal_theory] and_false
-attribute [minimal_theory] false_and
+attribute [minimal_theory high] and_true
```
shigoel, line 11: @alexkeizer A comment explaining why these rules are cheap, and thus can have a high priority, would be welcome.

alexkeizer: I added the following:
```
/-
Notice how both `and_true : ?p ∧ True = ?p` and `and_self : ?p ∧ ?p = ?p` may
be attempted by `simp` on a goal of the shape `_ ∧ True`, as they are both
a match where the discrimination tree is concerned.

However, assuming the left conjunct is not def-eq to `True`, an attempt of
`and_self` will fail to unify, which causes a fallback to `and_true`.
For the latter simp lemma, on the other hand, we know that if the discrimination
tree gives a match, then the lemma should be applicable.
This is because the variable `?p` is used only once in the pattern (i.e.,
the pattern is linear), and the first unification of an unassigned metavariable
is always successful.

Another possibly expensive lemma is
  `and_iff_left_of_imp : {a b : Prop} (ha : a → b) : a ∧ b ↔ a`
The pattern of this lemma (`?a ∧ ?b`) is perfectly linear, but to apply such
a lemma, `simp` first has to discharge the `(ha : a → b)`, which might fail,
and might be expensive.

Thus, the rationale we follow is that only linear, side-condition free lemmas
get the `high` priority, and everything else gets the default prioriry.
This ensures that `and_true` gets tried before `and_self` or
`and_iff_left_of_imp`
-/
```

I then also went through the rest of the file and mechanically applied this heuristic.

This seems to not have had much of an impact on the heartbeat usage, but still feels like morally the right thing to do. Experiments do show that some part of the LNSym standard simpsets is rather expensive, even if we don't use anything.

shigoel (approved):
Thanks, @alexkeizer!

PR #149, opened 2024-09-10
chore: yoink out common code from MaxTandem into separate files.

bollu:
This will be reused in the proof of `Memcpy`, and we thus perform this refactor first.


Arm/Syntax.lean
```diff
@@ -0,0 +1,17 @@
+import Arm.State
+/-
```
pennyannn, line 2: Nitpick: import is before header.

Proofs/Experiments/Max/MaxTandem.lean
```diff
@@ -8,12 +8,17 @@ computes the maximum of two numbers.
 -/
 import Arm
 import Tactics.CSE
+import Tactics.Rename
```
pennyannn, line 11: Duplicated `Tactics.Rename` imports

Tactics.lean
```diff
@@ -7,3 +7,6 @@ Author(s): Shilpi Goel
 -- Import modules here that should be built as part of the library.
 import «Tactics».Sym
 import «Tactics».StepThms
+import «Tactics».Rename
```
pennyannn, line 10: I'm wondering if these tactics about names are related and therefore could be put in the same file.

bollu: While I see the appeal, I prefer this approach, where it's stupidly easy to find out where a certain tactic is implemented.

pennyannn (commented):
LGTM

Arm/Syntax.lean
```diff
@@ -0,0 +1,16 @@
...
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Siddharth Bhat
+
+Provide convenient syntax for writing down state manipulation in ARM programs.
```
shigoel, line 6:
```suggestion
Provide convenient syntax for writing down state manipulation in Arm programs.
```

Arm/Syntax.lean
```diff
@@ -0,0 +1,16 @@
...
+-/
+import Arm.State
+
+namespace ArmStateNotation
+
+/-! We build a notation for `read_mem_bytes $n $base $s` as `$s[$base, $n]` -/
```
shigoel, line 12: Question: does this notation also show up in the goals? Or do we still see `read_mem_bytes...` there?

Tactics/ClearNamed.lean
```diff
@@ -0,0 +1,18 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Siddharth Bhat
+-/
+import Lean
+
+open Lean Elab Meta Tactic in
+/-- `clear_named [n₁, n₂, ...]` clear all inaccessible names that have
+*any* one of `n₁, n2, ...` as prefixes. -/
```
shigoel, line 10:
```suggestion
*any* one of `n₁, n₂, ...` as prefixes. -/
```

shigoel:
LGTM, minor nits!

Arm/Syntax.lean
bollu: I don't think we'll see it in the goals, since I wrote it as `syntax` plus `macro_rules` instead of `notation`, precisely because I wasn't sure if people want to see it in the goal state.

bollu:
@shigoel addressed nits!

shigoel (approved):

shigoel:
@bollu Please feel free to hit merge when CI checks pass. Thx!

Issue #155, opened 2024-09-16
Cache save failed in CI


shigoel:
CI can sometimes fail, though locally the build succeeds, likely due to a bad cache save. See the following logs for details.

In the first run, I'd switched ELFSage's Git URL to [my fork](https://github.com/shigoel/ELFSage/tree/main). 
https://github.com/leanprover/LNSym/actions/runs/10816223174

The second run is when I merged `main` into the branch.
https://github.com/leanprover/LNSym/actions/runs/10887193356/job/30208908292

For now, we have set `use-github-cache: false` under `leanprover/lean-action@v1`, which solves the issue.

Note that this scenario would need to be recreated -- the PR has been merged (though the branch is not yet deleted).
https://github.com/leanprover/LNSym/pull/151

shigoel:
Note that in both of the fails above, the cache save failure in on a MacOS runner.

PR #156, opened 2024-09-16
Remove explicit caching of .lake folder; enable use of Github cache

shigoel:
### Issues:

Hopefully, resolves https://github.com/leanprover/LNSym/issues/155

### Description:

Avoid caching fails in CI

### Testing:

`make all` ran locally

### License:

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -31,14 +31,7 @@ jobs:
       with:
         auto-config: false
         use-mathlib-cache: false
-        use-github-cache: false
-
-    - name: Cache `.lake` folder
-      id: cache-lake
-      uses: actions/cache@v4
-      with:
-        path: .lake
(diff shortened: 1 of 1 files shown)


bollu:
LGTM. As a note, we're doing this since I had hypothesized the following failure mode:

We have two caches:

```
    - uses: leanprover/lean-action@v1
      with:
        auto-config: false
        use-mathlib-cache: false
        use-github-cache: true

    - name: Cache `.lake` folder
      id: cache-lake
      uses: actions/cache@v4
      with:
        path: .lake
        key: ${{ runner.os }}-lake
```

I feel we should delete the hand-written `Cache .lake folder` rule, and only keep `leanprover/lean-action`'s `use-github-cache: true` rule. I could very well imagine that the latter cache somehow overwrote files from the former's cache, thereby invalidating the .lake state!

PR #157, opened 2024-09-16
chore: add extract goal from Mathlib

bollu:
This incurs a dependency onto `Batteries`.


bollu:
I'm going to close this, since we decided to not go through with this, and request upstream instead to have `extract_goal`.

shigoel:
@bollu Thanks, Sid.

PR #158, opened 2024-09-16
Use structural recursion and rfl

pennyannn:
### Description:

This PR changes well-founded recursion to structural recursion in several functions and this allows us to use `rfl` for the tests that associate with these functions.

Relevant PR: https://github.com/leanprover/LNSym/pull/151

### Testing:

### License:

### DIFF
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -294,7 +294,7 @@ abbrev partInstall (hi lo : Nat) (val : BitVec (hi - lo + 1)) (x : BitVec n): Bi
   let x_with_hole := x &&& mask_with_hole
   x_with_hole ||| val_aligned
 
-example : (partInstall 3 0 0xC#4 0xAB0D#16 = 0xAB0C#16) := by native_decide
+example : (partInstall 3 0 0xC#4 0xAB0D#16 = 0xAB0C#16) := by rfl
 
 def flattenTR {n : Nat} (xs : List (BitVec n)) (i : Nat)
   (acc : BitVec len) (H : n > 0) : BitVec len :=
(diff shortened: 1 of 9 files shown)


pennyannn:
Wow, apparently there have been great improvements in Lean performance. Lat time I tried `decide`/`rfl` for some of these tests, it takes minutes to finish. Now it finishes almost instantly! This helped removing several `native_decide`s.

Specs/GCMV8.lean
```diff
@@ -204,7 +201,7 @@ example :  GCMInitV8 0x66e94bd4ef8a2c3b884cfa59ca342b2e#128 ==
     0x4af32418184aee1eec87cfb0e19d1c4e#128,
     0xf109e6e0b31d1eee7d1998bcfc545474#128,
     0x7498729da40cd2808c107e5c4f494a9a#128,
-    0xa47c653dfbeac924d0e417a05fe61ba4#128 ] := by native_decide
+    0xa47c653dfbeac924d0e417a05fe61ba4#128 ] := by rfl
```
shigoel, line 204: hurray!

shigoel (dismissed):
LGTM!

Arm/BitVec.lean
```diff
@@ -308,37 +308,35 @@ def flattenTR {n : Nat} (xs : List (BitVec n)) (i : Nat)
...
-  -- (FIXME) With leanprover/lean4:nightly-2024-08-29, just `rfl` sufficed here.
-  simp [reverse, reverse.reverseTR]
-  rfl
+    match i with
+    | 0 => acc
+    | j + 1 =>
+      have h1 : i - 1 - (i - 1) + 1 = 1 := by omega
+      let xi : BitVec 1 := BitVec.cast h1 $ extractLsb (i - 1) (i - 1) x
+      have h2 : 1 = (n - i) - (n - i) + 1 := by omega
+      let acc := BitVec.partInstall (n - i) (n - i) (BitVec.cast h2 xi) acc
```
alexkeizer, line 317: You cast `xi` twice, that seems a bit redundant. We could either merge the two `have`s into one, or if we don't mind a little inference, the following should work:
```suggestion
      let xi : BitVec 1 := extractLsb (i - 1) (i - 1) x
      let acc := BitVec.partInstall (n - i) (n - i) (xi.cast (by omega)) acc
```

Arm/BitVec.lean
```diff
@@ -308,37 +308,35 @@ def flattenTR {n : Nat} (xs : List (BitVec n)) (i : Nat)
...
-    if i < n/e then
-      let lo := i * e
+    match i with
+    | 0 => acc
+    | j + 1 =>
+      let lo := (n / e - i) * e
       let hi := lo + e - 1
       have h₀ : hi - lo + 1 = e := by simp only [hi, lo]; omega
       let part : BitVec e := BitVec.cast h₀ (extractLsb hi lo x)
       let newacc := part :: acc
```
alexkeizer, line 334: (This is unrelated to the main change, so feel free to ignore)

Generally, using `extractLsb'` is nicer than `extractLsb`, because you specify the length directly and can thus generally avoid casts. 
```suggestion
      let part : BitVec e := extractLsb' lo e x
      let newacc := part :: acc
```

(do double check if the start should be `lo` or `lo +/- 1`, I always forget if there's any inclusive/exclusive interval shenanigans going on with these)

Specs/GCMV8.lean
```diff
@@ -188,10 +187,8 @@ def GCMInitV8 (H : BitVec 128) : (List (BitVec 128)) :=
   [H0_rev, H1, H2_rev, H3_rev, H4, H5_rev, H6_rev,
    H7, H8_rev, H9_rev, H10, H11_rev]
 
--- set_option profiler true in
--- set_option maxRecDepth 20000 in
--- set_option maxHeartbeats 2000000 in
--- unseal pmod.pmodTR degree.degreeTR reverse.reverseTR pmult.pmultTR Nat.bitwise in
+set_option maxRecDepth 20000 in
```
alexkeizer, line 190: This `maxRecDepth` increase is a bit scary, do we need it? Increasing the maxRecDepth can cause Lean to rather ungracefully crash with stack-overflows

Arm/BitVec.lean
pennyannn: Thanks for the suggestion. Several functions are simplified this way.

Specs/GCMV8.lean
pennyannn: I have tried and was able to reduce depth to 8000. I was also able to remove the maxHeartbeat. Unfortunately these maxRecDepth could not be fully removed.

Arm/BitVec.lean
```diff
@@ -294,7 +294,7 @@ abbrev partInstall (hi lo : Nat) (val : BitVec (hi - lo + 1)) (x : BitVec n): Bi
   let x_with_hole := x &&& mask_with_hole
   x_with_hole ||| val_aligned
 
-example : (partInstall 3 0 0xC#4 0xAB0D#16 = 0xAB0C#16) := by native_decide
+example : (partInstall 3 0 0xC#4 0xAB0D#16 = 0xAB0C#16) := by rfl
```
alexkeizer, line 297:
```suggestion
example : (partInstall 3 0 0xC#4 0xAB0D#16 = 0xAB0C#16) := rfl
```

(minor nit) If a proof is just `rfl`, we prefer term-mode over tactic mode

Arm/BitVec.lean
```diff
@@ -308,37 +308,31 @@ def flattenTR {n : Nat} (xs : List (BitVec n)) (i : Nat)
...
-  rfl
+    match i with
+    | 0 => acc
+    | j + 1 =>
+      let xi : BitVec 1 := extractLsb' (i - 1) 1 x
+      let acc := BitVec.partInstall (n - i) (n - i) (xi.cast (by omega)) acc
+      reverseTR x j acc
+  reverseTR x n $ BitVec.zero n
+
+example : reverse 0b11101#5 = 0b10111#5 := by rfl
```
alexkeizer, line 319:
```suggestion
example : reverse 0b11101#5 = 0b10111#5 := rfl
```
And a bunch more

Specs/GCMV8.lean
alexkeizer: I guess it's fine to have them, if it can't be avoided. Better than having `native_decide`, anyway

alexkeizer (commented):
LGTM

Arm/BitVec.lean
pennyannn: I've fixed a bunch of these in the repo, see new commit.

shigoel (approved):
Thanks, @pennyannn and @alexkeizer!

PR #159, opened 2024-09-17
Popcount32 functional correctness and (non)effects characterization

shigoel:
### Description:

Prove that the popcount32 program correctly computes the bitcount of its input, and prove that certain state components were left unmodified by this program.

### Testing:

`make all` ran locally.

### License:

### DIFF
--- a/Proofs/Popcount32.lean
+++ b/Proofs/Popcount32.lean
@@ -5,6 +5,8 @@ Author(s): Nathan Wetzler, Shilpi Goel, Alex Keizer
 -/
 import Arm.Exec
 import Arm.Util
+import Arm.Syntax
+import Arm.Memory.SeparateAutomation
 import Tactics.Sym
 import Tactics.Aggregate
 import Tactics.StepThms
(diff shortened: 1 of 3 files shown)


alexkeizer (commented):
@shigoel the non-effects proof now goes through fully automatically. The key was setting the `contextual` of simp, so that the non-equality hypotheses were automatically added to the simpset.

Proofs/Popcount32.lean
```diff
@@ -68,44 +71,57 @@ def popcount32_program : Program :=
...
---        clear h_step_21 h_step_22 h_step_23 h_step_24;
---        clear h_step_25 h_step_26)
---   -- Final Steps
---   unfold run at h_run
---   subst s_final
---   unfold popcount32_spec
---   sorry
+  -- Final Steps
+  -- Split all the Ands in the conclusion.
+  repeat (any_goals apply And.intro)
```
alexkeizer, line 93: `repeat (any_goals _)` looks a bit funny, is `repeat` by itself not enough?

Proofs/Popcount32.lean
```diff
@@ -68,44 +71,57 @@ def popcount32_program : Program :=
...
+  · simp only [popcount32_spec,
+               fst_AddWithCarry_eq_add,
+               fst_AddWithCarry_eq_sub_neg]
+    repeat (simp only [popcount32_spec_rec])
```
alexkeizer, line 97: Here, too, `repeat (simp _)` looks weird. We only have one goal, right? So `simp` should already repeatedly apply the rewrite

alexkeizer (commented):
LGTM, modulo two minor comments

Proofs/Popcount32.lean
alexkeizer: @bollu told me that, no, `repeat` will only repeat on the main goal.

However, `repeat'` does work on all goals, and should work here

shigoel: Hah, thanks!

PR #161, opened 2024-09-17
feat: add option for a custom `simp` config to aggregate tactic

alexkeizer:
### Description:

Adds an option to configure the `simp` config used in `sym_aggregate`.
Also, enables `contextual` by defaul, which makes it powerful enough to prove non-effect goals like those found in #159.

Note that we *don't* enable contextual by default in the aggregation post-processing step of `sym_n`, since it seemed to trigger a timeout there.

### Testing:

### License:

### DIFF
--- a/Tactics/Aggregate.lean
+++ b/Tactics/Aggregate.lean
@@ -11,15 +11,25 @@ open Lean Meta Elab.Tactic
 
 namespace Sym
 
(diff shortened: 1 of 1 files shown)


alexkeizer:
We decided to merge this together with the popcount proof in #159, to avoid the merge song-and-dance

PR #163, opened 2024-09-18
feat: bitvec lemmas for `memcpy` [1/? memcpy]

bollu:
We add lemmas about (non) overflow, as well as theorems that show that not-equals-zero iff equals-one


shigoel (approved):
LGTM!

PR #164, opened 2024-09-18
feat: add lemma: Z flag = 1 iff value = 0. [2/? memcpy]

bollu:


Arm/Insts/Common.lean
```diff
@@ -145,6 +145,18 @@ theorem sle_iff_not_n_eq_v_and_z_eq_0_32 (x y : BitVec 32) :
   · bv_decide
   · bv_decide
 
+/--
+`x = 0` iff `Z = 0`.
+This is implemented by testing whether `x + (-1) + 1 = 0`
+-/
+theorem zero_iff_z_eq_one (x : BitVec 64) :
```
shigoel, line 152: Aren't we adding all `AddWithCarry` lemmas to a simpset? Or is that under @alexkeizer's purview?

bollu: @shigoel I'm currently being conservative, since I've found that proofs sometimes break, so I'd like to split such changes into two PRs. I take on the responsibility of sending the follow-up PR myself.

shigoel (approved):

PR #166, opened 2024-09-18
chore: add q0..4 accessors for registers [4/? memcpy]

bollu:


Arm/State.lean
```diff
@@ -336,7 +336,7 @@ def r (fld : StateField) (s : ArmState) : (state_value fld) :=
 /-!
 
 We define helpers for reading and writing registers on the `ArmState` with the colloquial
-names. For example, the stack pointer (`sp`) refers to register 31. 
+names. For example, the stack pointer (`sp`) refers to register 31.
 These mnemonics make it much easier to read and write theorems about assembly programs.
```
shigoel, line 340: Wondering if we ought to move these into Arm/Syntax.lean instead. Thoughts?

shigoel, line 340: Wondering if we ought to move these into Arm/Syntax.lean instead. Thoughts?

bollu: I think I'd prefer to keep it in `State`, leaving `Syntax` to be for syntax sugar :)

shigoel (approved):

PR #167, opened 2024-09-18
feat: add tactic to skip proof checking during development. [PR 5/? for memcpy]

bollu:
Surrounding a tactic with `skip_proof` will skip running the tactic and close the goal with `skipProofAx`.
We only skip proofs `skip_proof.skip` is set to `true`, and we emit a warning to remind the user that proofs are being skipped.

This is to be used, during development, with expensive tactics such as `omega`, `bv_omega`, `bv_decide`, `simp_mem`.

This allows one to run `have expensive_hyp : ... := by omega`, check that this can be proven,
and to then replace this with `have expensive_hyp : ... := by skip_proof omega`,
which will skip the use of `omega`, and instead of `skipProoxAx`, ensuring Lean stays interactive. 

This makes `skip_proof p` morally different from a `sorry`:
- It still retains the proof `p`. This signals the intent of a completed proof that has been skipped for performance.
- It uses a different axiom (`skipProofAx` instead of `sorryAx`).
  This allows `#print axioms in thm` to distinguish between skipped proofs and genuine `sorry`s.

### DIFF
--- a/Tactics/SkipProof.lean
+++ b/Tactics/SkipProof.lean
@@ -1,0 +1,42 @@
+import Lean
+
+open Lean Meta Elab Tactic
+
+namespace SkipProof
+
+register_option skip_proof.skip : Bool := {
+  defValue := False
+  descr := "If `skip` is true, `skip_proof p` skips running `p` and closes the goal with `skipProofAx`."
+}
+
+/--
+Surrounding a tactic with `skip_proof` will skip running the tactic and close the goal with `skipProofAx`.
+We only skip proofs `skip_proof.skip` is set to `true`,
+and we emit a warning to remind the user that proofs are being skipped.
+
+This is to be used, during development, with expensive tactics such as `omega`, `bv_omega`, `bv_decide`, `simp_mem`.
+This allows one to run `have expensive_hyp : ... := by omega`, check that this can be proven,
+and to then replace this with `have expensive_hyp : ... := by skip_proof omega`,
+which will skip the use of `omega`, and instead of `skipProoxAx`, ensuring Lean stays interactive.
+
+`skip_proof p` signals that when `set_option skip_proof.skip false` is set, the proof `p` *will go through*.
+This makes `skip_proof p` morally different from a `sorry`:
+- It still retains the proof `p`. This signals the intent of a completed proof that has been skipped for performance.
+- It uses a different axiom (`skipProofAx` instead of `sorryAx`).
+  This allows `#print axioms in thm` to distinguish between skipped proofs and genuine `sorry`s.
+-/
+syntax (name := skip_proof) "skip_proof" tactic : tactic
+axiom skipProofAx {α : Sort _} : α
+
+@[tactic skip_proof]
+def skipProof : Tactic := fun
+  | `(tactic| skip_proof $x) => do
+    let opts ← Lean.getBoolOption `skip_proof.skip
+    if opts then
+      logWarningAt x "Skipped by `skip_proof`"
+      evalTactic (← `(tactic| exact skipProofAx))
+    else
+      evalTactic x
+  | _ => throwUnsupportedSyntax
+
+end SkipProof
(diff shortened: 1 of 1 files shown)


Tactics/SkipProof.lean
```diff
@@ -0,0 +1,42 @@
+import Lean
```
shigoel, line 1: Missing license and copyright header

Tactics/SkipProof.lean
```diff
@@ -0,0 +1,42 @@
...
+namespace SkipProof
+
+register_option skip_proof.skip : Bool := {
+  defValue := False
+  descr := "If `skip` is true, `skip_proof p` skips running `p` and closes the goal with `skipProofAx`."
+}
+
+/--
+Surrounding a tactic with `skip_proof` will skip running the tactic and close the goal with `skipProofAx`.
+We only skip proofs `skip_proof.skip` is set to `true`,
```
shigoel, line 14:
```suggestion
We only skip proofs if `skip_proof.skip` is set to `true`,
```

Tactics/SkipProof.lean
```diff
@@ -0,0 +1,42 @@
...
+and we emit a warning to remind the user that proofs are being skipped.
+
+This is to be used, during development, with expensive tactics such as `omega`, `bv_omega`, `bv_decide`, `simp_mem`.
+This allows one to run `have expensive_hyp : ... := by omega`, check that this can be proven,
+and to then replace this with `have expensive_hyp : ... := by skip_proof omega`,
+which will skip the use of `omega`, and instead of `skipProoxAx`, ensuring Lean stays interactive.
```
shigoel, line 20: Wording went wonky here.. "and instead of `skipProoxAx`..."

Tactics/SkipProof.lean
```diff
@@ -0,0 +1,42 @@
...
+
+`skip_proof p` signals that when `set_option skip_proof.skip false` is set, the proof `p` *will go through*.
+This makes `skip_proof p` morally different from a `sorry`:
+- It still retains the proof `p`. This signals the intent of a completed proof that has been skipped for performance.
```
shigoel, line 24: Skipped for performance during interactive use, I'd emphasize

shigoel:
Good to think about upstreaming this to core, if Lean folks would have it.

Tactics/SkipProof.lean
bollu: Fixed.

Tactics/SkipProof.lean
bollu: Done.

Tactics/SkipProof.lean
bollu: Fixed!

Tactics/SkipProof.lean
bollu: Done!

bollu:
Pushed.

Tactics/SkipProof.lean
```diff
@@ -1,3 +1,10 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author(s): Siddharth Bhat
+
+Tactic to skip proof checking for interactive use.
+-/
```
shigoel, line 7: Could you have this as a separate comment from the copyright header?

bollu: @shigoel I'm not sure why, isn't this how we do it always? https://github.com/leanprover/lean4/blob/master/src/Init/Meta.lean#L6

bollu: I'm not sure that file is the right example, since there's no file-level comment in it :D I guess a file such as this: https://github.com/leanprover/LNSym/blob/main/Correctness/Correctness.lean#L7-L15 shows the LNSym style. Happy to oblige.

PR #169, opened 2024-09-19
feat: increase simp_mem power by trying to close goals with omega [6/? memcpy]

bollu:
This lets us close goals where the hypotheses are about memory, and the goal is an LIA goal. For example:

```
h : mem_separate' a 10 b 10
hab : a < b
-----------
a + 5 < b
```

### DIFF
--- a/Arm/Memory/SeparateAutomation.lean
+++ b/Arm/Memory/SeparateAutomation.lean
@@ -702,7 +702,6 @@ For example, `mem_legal.of_omega` is a function of type:
 class OmegaReducible (α : Type) where
   reduceToOmega : α → Expr
 
-
 /--
 info: mem_legal'.of_omega {n : Nat} {a : BitVec 64} (h : a.toNat + n ≤ 2 ^ 64) : mem_legal' a n
 -/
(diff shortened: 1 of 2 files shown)


Arm/Memory/SeparateAutomation.lean
```diff
@@ -962,12 +961,29 @@ partial def SimpMemM.closeGoal (g : MVarId) (hyps : Array Hypothesis) : SimpMemM
...
-        if let .some proof ←  proveWithOmega? e hyps then
+        if let .some proof ← proveWithOmega? e hyps then
           g.assign proof.h
     if let .some e := MemSeparateProp.ofExpr? gt then
       withTraceNode m!"Matched on ⊢ {e}. Proving..." do
-        if let .some proof ←  proveWithOmega? e hyps then
+        if let .some proof ← proveWithOmega? e hyps then
           g.assign proof.h
+
+    withTraceNode m!"Unknown memory expression kind ⊢ {gt}. Trying a reduction to omega..." do
```
shigoel, line 971: `omega` can be expensive, so let's build an option to turn this off when we know it's not going to help

bollu:
Flags added!

shigoel (approved):

PR #168, opened 2024-09-18
refactor: cleanup popcount proof

alexkeizer:
### Description:

We've cleaned up the popcount proof by:
* extending `sym_aggregate` to also search for memory effect hypotheses to simplify, and
* changing the semantics of `add_sub_immediate` to avoid pattern-matching. It turns out that pattern-matching on the result of an `if` (e.g., `let (a, b) := if x = 0 then (foo, bar) else (bar, foo)`) get's reduced to a `Decidable.rec` by `simp only`, which then gets stuck. By rephrasing it in terms of projections, we avoid this, and get clean simplified semantics.
* In the same vein, we also add existing simplification rules for `AddWithCarry` to the `bitvec_rules` simpset

### Testing:

`make all` succeeds.

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #171, opened 2024-09-19
chore: rewrite semantics to not have pattern match on structure

bollu:
### Issues:

This works around https://github.com/leanprover/lean4/issues/5388, and removes uses of `simp (config := { ground := true }) ..`. This makes our proof states cleaner, since we no longer see proof states involving `Decidable.rec ...`. Peeled from #168 

### Testing:

`make all` succeeded, conformance succeeded.

### License:

### DIFF
--- a/Arm/Insts/DPI/Add_sub_imm.lean
+++ b/Arm/Insts/DPI/Add_sub_imm.lean
@@ -24,19 +24,21 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
                         else
                           imm <<< 12
     let operand1      := read_gpr datasize inst.Rn s
-    let (carry_in, operand2)
+    let carryInAndOperand2
                       := if sub_op then
                           (1#1, ~~~imm)
                         else
                           (0#1, imm)
+    let carry := carryInAndOperand2.fst
+    let operand2 := carryInAndOperand2.snd
     let operand2         := BitVec.zeroExtend datasize operand2
-    let (result, pstate) := AddWithCarry operand1 operand2 carry_in
+    let resultAndPState := AddWithCarry operand1 operand2 carry
     -- State Updates
     let s'            := write_pc ((read_pc s) + 4#64) s
-    let s'            := if setflags then write_pstate pstate s' else s'
+    let s'            := if setflags then write_pstate resultAndPState.snd s' else s'
     let s'            := if inst.Rd = 31#5 ∧ ¬ setflags
-                         then write_gpr datasize inst.Rd result s'
-                         else write_gpr_zr datasize inst.Rd result s'
+                         then write_gpr datasize inst.Rd resultAndPState.fst s'
+                         else write_gpr_zr datasize inst.Rd resultAndPState.fst s'
     s'
 
 ----------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


shigoel:
BTW, `AddWithCarry` is also used here:
```
./Insts/DPR/Add_sub_shifted_reg.lean:    let (result, pstate) := AddWithCarry operand1 operand2 carry
./Insts/DPR/Add_sub_carry.lean:    let (result, pstate) := AddWithCarry operand1 operand2 (read_flag PFlag.C s)
```

Arm/Insts/DPI/Add_sub_imm.lean
```diff
@@ -24,19 +24,21 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
                         else
                           imm <<< 12
     let operand1      := read_gpr datasize inst.Rn s
-    let (carry_in, operand2)
+    let carryInAndOperand2
```
shigoel, line 27: Could we leave a note (and an issue) to remind us to restore this definition to the original form after https://github.com/leanprover/lean4/issues/5388 is fixed?

Arm/Insts/DPI/Add_sub_imm.lean
```diff
@@ -24,19 +24,21 @@ def exec_add_sub_imm (inst : Add_sub_imm_cls) (s : ArmState) : ArmState :=
...
                       := if sub_op then
                           (1#1, ~~~imm)
                         else
                           (0#1, imm)
+    let carry := carryInAndOperand2.fst
+    let operand2 := carryInAndOperand2.snd
     let operand2         := BitVec.zeroExtend datasize operand2
-    let (result, pstate) := AddWithCarry operand1 operand2 carry_in
+    let resultAndPState := AddWithCarry operand1 operand2 carry
```
alexkeizer, line 35: This pattern-match should be fine, right? IIUC, it was only matching on an if that caused problems

bollu:
> BTW, `AddWithCarry` is also used here:
> 
> ```
> ./Insts/DPR/Add_sub_shifted_reg.lean:    let (result, pstate) := AddWithCarry operand1 operand2 carry
> ./Insts/DPR/Add_sub_carry.lean:    let (result, pstate) := AddWithCarry operand1 operand2 (read_flag PFlag.C s)
> ```


@shigoel these are okay, because their arguments don't come from a `if...then...else`. Consider the diff:

```
-    let (carry_in, operand2)
+    let carryInAndOperand2
                       := if sub_op then
                           (1#1, ~~~imm)
                         else
                           (0#1, imm)
+    let carry := carryInAndOperand2.fst
+    let operand2 := carryInAndOperand2.snd
     let operand2         := BitVec.zeroExtend datasize operand2
-    let (result, pstate) := AddWithCarry operand1 operand2 carry_in
+    let (result, pstate) := AddWithCarry operand1 operand2 carry
```

The issue is that we were pattern matching to extract `(carry_in, operand2)` from the `if ... then ... else` which elaborated poorly. The culprit is this pattern match, not so much the use of `AddWithCarry`. `AddWithCarry` would then try to use the `carry`, but any function that used `carry` would have lead to a problem.

Arm/Insts/DPI/Add_sub_imm.lean
bollu: Done!

Arm/Insts/DPI/Add_sub_imm.lean
bollu: Fixed, undone.

bollu:
@shigoel updated.

shigoel (approved):

PR #177, opened 2024-09-21
feat: add option in `LnSymSimpContext` to add arbitrary expressions to the simpset

alexkeizer:
### Description:

This allows us to `simp` with expressions that aren't just fvars or constants.

### Testing:

### License:


shigoel:
@alexkeizer Could you give a usage example here?

alexkeizer:
This will be used when moving to intermediate state aggregation, where we won't necessarily add all the expressions in an `AxEffect` to the environment, but we still want to be able to simplify along them.

shigoel:
Interesting, thanks, @alexkeizer! Looking forward to seeing this in action.

shigoel (approved):

PR #180, opened 2024-09-23
refactor: incorporate an AxEffects field in SymContext, introduce `SymM` monad for `SymContext` state

alexkeizer:
### Description:

This turned into a rather big tech-debt removal PR. The primary focus is the removal of the duplication we had between `SymContext` tracking names of hypotheses, and `AxEffects` tracking `Expr`s for those same hypotheses.

- Added an `effects : AxEffects` field to `SymContext`, which stores the `AxEffects` for a single (non-aggregated!) step
- To make live easier, we introduce a `SymM` monad, so that we don't have to project out to the effect field every single every time (credits/blame go to @bollu for showing me the `MonadStateOf` trick).
- This allowed us to remove a bunch of fields of `SymContext` which had duplicates in `AxEffects`
- Also, it allowed us to move reflection of the PC out of `explodeSteps` and into `prepareForNextStep` (which was previously called `SymContext.next`),
- Finally, we extract `ensureAtMostRunSteps` and `assertStepTheoremsGenerated` functions out of the main body of `sym_n`

### Testing:

### License:

### DIFF
--- a/Tactics/Common.lean
+++ b/Tactics/Common.lean
@@ -223,9 +223,9 @@ def findLocalDeclOfType? (expectedType : Expr) : MetaM (Option LocalDecl) := do
     --    the local context, so we can safely pass it to `get!`
 
 def findLocalDeclOfTypeOrError (expectedType : Expr) : MetaM LocalDecl := do
-    let some name ← findLocalDeclOfType? expectedType
+    let some decl ← findLocalDeclOfType? expectedType
       | throwError "Failed to find a local hypothesis of type {expectedType}"
-    return name
+    return decl
 
 /-- `findProgramHyp` searches the local context for an hypothesis of type
   `state.program = ?concreteProgram`,
--- a/Tactics/Common.lean
+++ b/Tactics/Common.lean
@@ -269,3 +269,12 @@ def traceHeartbeats (cls : Name) (header : Option String := none) :
   let percent ← heartbeatsPercent
   trace cls fun _ =>
     m!"{header}used {heartbeats} heartbeats ({percent}% of maximum)"
+
+/-! ## `withMainContext'` -/
+
+variable {m} [Monad m] [MonadLiftT TacticM m] [MonadControlT MetaM m]
+/-- Execute `x` using the main goal local context and instances.
+
+Unlike the standard `withMainContext`, `x` may live in a generic monad `m`. -/
+def withMainContext' (x : m α) : m α := do
+  (← getMainGoal).withContext x
--- a/Tactics/Reflect/AxEffects.lean
+++ b/Tactics/Reflect/AxEffects.lean
@@ -82,6 +82,23 @@ structure AxEffects where
 
 namespace AxEffects
 
+/-! ## Monad getters -/
+
+section Monad
+variable {m} [Monad m] [MonadReaderOf AxEffects m]
+
+def getCurrentState       : m Expr := do return (← read).currentState
+def getInitialState       : m Expr := do return (← read).initialState
+def getNonEffectProof     : m Expr := do return (← read).nonEffectProof
+def getMemoryEffect       : m Expr := do return (← read).memoryEffect
+def getMemoryEffectProof  : m Expr := do return (← read).memoryEffectProof
+def getProgramProof       : m Expr := do return (← read).programProof
+
+def getStackAlignmentProof? : m (Option Expr) := do
+  return (← read).stackAlignmentProof?
+
+end Monad
+
 /-! ## Initial Reflected State -/
 
 /-- An initial `AxEffects` state which has no writes.
--- a/Tactics/Reflect/AxEffects.lean
+++ b/Tactics/Reflect/AxEffects.lean
@@ -377,6 +377,28 @@ private def assertIsDefEq (e expected : Expr) : MetaM Unit := do
     throwError "expected:\n  {expected}\nbut found:\n  {e}"
 
 /-- Given an expression `e : ArmState`,
+which is a sequence of `w`/`write_mem`s to `eff.currentState`,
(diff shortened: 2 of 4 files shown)


Tactics/Reflect/AxEffects.lean
```diff
@@ -359,6 +395,28 @@ private def assertIsDefEq (e expected : Expr) : MetaM Unit := do
   if !(←isDefEq e expected) then
     throwError "expected:\n  {expected}\nbut found:\n  {e}"
 
+/-- Given an expression `e : ArmState`,
+which is a sequence of `w`/`write_mem`s to `eff.currentState`,
+return an `AxEffects` where `e` is `currentState`.
+
+Note that as soon as an unsupported expression (e.g., an `if`) is encountered,
+the whole expression is taken to be the initial state,
```
shigoel, line 403: I know this line is from before, but it could be made more accurate. Right now, it seems as though this is an assumption but isn't an error thrown if the expressions isn't def-eq to the initial state?
```
the whole expression is taken to be the initial state,
```

Tactics/Sym.lean
```diff
@@ -42,25 +44,30 @@ macro "init_next_step" h_run:ident stepi_eq:ident sn:ident : tactic =>
...
-          $ctx.h_err_ident:ident)
-  ))
+def stepiTac (stepiEq : Expr) (hStep : Name) : SymReaderM Unit := fun ctx =>
+  withMainContext' do
+    let pc := (Nat.toDigits 16 ctx.pc.toNat).asString
+    --  ^^ The PC in hex
+    let stepLemma := Name.str ctx.program s!"stepi_eq_0x{pc}"
+    -- let stepLemma := Expr.const stepLemma []
+
+    let eff := ctx.effects
```
shigoel, line 59: No `evalTactic` here? Do we see a perf. improvement now?

Tactics/Sym.lean
```diff
@@ -309,18 +301,54 @@ def sym1 (c : SymContext) (whileTac : TSyntax `tactic) : TacticM SymContext :=
...
+  let pc := c.pc.toHexWithoutLeadingZeroes
+  let step_thm := Name.str c.program ("stepi_eq_0x" ++ pc)
+  try
+    let _ ← getConstInfo step_thm
+  catch err =>
+    throwErrorAt err.getRef "{err.toMessageData}\n
+Did you remember to generate step theorems with:
+  #genStepEqTheorems {c.program}
+
+Alternatively, are you sure that {pc} is a valid value for the program counter?"
```
shigoel, line 349: Could you also define "valid" here? Perhaps "concrete" is a better word, for now.

Tactics/Common.lean
```diff
@@ -269,3 +269,12 @@ def traceHeartbeats (cls : Name) (header : Option String := none) :
   let percent ← heartbeatsPercent
   trace cls fun _ =>
     m!"{header}used {heartbeats} heartbeats ({percent}% of maximum)"
+
+/-! ## `withMainContext'` -/
+
+variable {m} [Monad m] [MonadLiftT TacticM m] [MonadControlT MetaM m]
```
shigoel, line 275: Can we limit the scope of this `variable`? It's only used in `withMainContext'`, right?

Tactics/SymContext.lean
```diff
@@ -132,57 +148,87 @@ def findFromUserName (name : Name) : MetaM LocalDecl := do
...
-  else
-    { ctxt with
-      state_prefix := "s",
-      curr_state_number := 1 }
+  modifyThe SymContext fun ctxt =>
+    if let some currentStateNumber := tail.toNat? then
+      { ctxt with
+        state_prefix := (state.get? 0).getD 's' |>.toString,
+        currentStateNumber }
+    else
```
shigoel, line 228: Let's print a warning here that tells the user that `s1` is the state var we're running with

Tactics/Sym.lean
alexkeizer: In this case, valid means whether the program has an instruction at the specified PC. We can actually detect this information, so I've added a separate, more specific, error for that case

Tactics/Sym.lean
alexkeizer: Indeed, I got rid of the evalTactic so I could use Exprs, and we should see a performance improvement (both from circumventing the evalTactic, and from not having to look up names in the context as often)

Tactics/Common.lean
alexkeizer: Yes

Tactics/Sym.lean
shigoel: Thanks, Alex. That'd be awesome!

Tactics/Reflect/AxEffects.lean
alexkeizer: You're right, this line is only true for `fromExpr`, not for `updateExpr` (we check that the expression is def-eq the the previous *currentState*, but same difference)

Tactics/Sym.lean
shigoel: @alexkeizer  Need this branch to lie around then for perf., even after it's merged?

alexkeizer: If it's merged, I can just use the commit ID from main, so no need

Tactics/SymContext.lean
alexkeizer: Sure, I've added a warning

bollu:
@alexkeizer and I minimized this, and will be sending an issue upstream soon. In the meantime, we also have a workaround so it's all good :)

alexkeizer:
I pushed the workaround, @shigoel this is once more ready for review (and at the top of the priority queue from my side)

shigoel (approved):
Thanks!

PR #183, opened 2024-09-23
feat: benchmark setup

alexkeizer:
### Description:

Add a basic benchmark setup:
- `benchmark` command, which has a theorem-like syntax, but doesn't get added to the environment, and runs the proof 5 times, reporting both the individual and average runtime,
- similarly, a `logHeartbeats` tactic combinator, which measures the heartbeats used by an arbitrary tactic.
- Add a benchmark folder, with files for SHA512 runs with: 150, 225, and 400 steps (which are the numbers I've been reporting so far)
- Adds a `make benchmark` Make-target, to build the aforementioned files. Crucially, this target is *not* part of `make all`, for all our sanity's sake (running 400 steps of SHA512 for 5 times takes a while)

### Testing:

### License:


alexkeizer:
@shigoel, @bollu: do we want to run benchmarks in CI? It could be nice to gather data automatically, but I worry a bit that the data might be unreliable (since we have little to no control over what else is running on the CI worker), and that it might slow down CI too much (just doing the 400 step benchmark alone already takes 10 minutes)

Benchmarks/Command.lean
```diff
@@ -0,0 +1,47 @@
...
+
+open Lean Parser.Command Elab.Command
+
+elab "benchmark" id:ident declSig:optDeclSig val:declVal : command => do
+  let stx ← `(command| example $declSig:optDeclSig $val:declVal )
+
+  let n := 5
+  let mut runTimes := #[]
+  let mut totalRunTime := 0
+  for _ in List.range n do
```
bollu, line 16:
```suggestion
  for _ in [0:n] do
```

why not use range syntax? :)

Benchmarks/Command.lean
```diff
@@ -0,0 +1,47 @@
...
+
+  let avg := totalRunTime.toFloat / n.toFloat / 1000
+  logInfo m!"\
+{id}:
+  average runtime over {n} runs:
+    {avg}s
+
+  indidividual runtimes:
+    {runTimes}
+"
```
bollu, line 32:
```suggestion
  let avg := totalRunTime.toFloat / n.toFloat / 1000
  let geomean := (totalRunTime.toFloat.pow (1.0 / n.toFloat)) / 1000.0
  logInfo m!"\
{id}:
  average runtime over {n} runs:
    {avg}s
  geomean over {n} runs:
    {geomean}s

  indidividual runtimes:
    {runTimes}
"
```

Benchmarks/Command.lean
alexkeizer: At the risk of losing my street cred: because I wasn't aware of range syntax 🫢

Benchmarks/Command.lean
bollu: Ideally, we would also track variance, but I'm happy to do that in a follow up PR.

shigoel:
@alexkeizer I like the idea of benchmarking automatically in CI (even with all the caveats that we can't control what else is running on the worker). Maybe we can separate the benchmarking job from others so that folks don't have to wait for the benchmarking job to finish before hitting merge, etc. if they don't want to.

shigoel:
@alexkeizer Also, could you please update the README to describe the new `benchmark` target?

bollu:
A separate build job makes sense to me, and PR LGTM!

shigoel:
@alexkeizer FYI: incoming merge conflicts here.

alexkeizer:
@shigoel I've added a line to the README, and made benchmarking part of the CI (on the ubuntu runner only, figured that'd be enough). This should now be ready to merge

shigoel (approved):
Thanks!

PR #187, opened 2024-09-24
refactor: rename `Reflect` folder to `Sym`, move `SymContext` to `Sym/Context`

alexkeizer:
### Description:

Stacked on #182.

The new name makes more sense, since this is where I've been storing various components of the `sym_n` tactic implementation, even if they're not really doing reflection

### Testing:

What tests have been run? Did `make all` succeed for your changes? Was
conformance testing successful on an Aarch64 machine? Yes

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #189, opened 2024-09-25
feat: LCtxSearch abstraction to search the local context in a single pass

alexkeizer:
### Description:

The main purpose of this definition is to build an abstraction around searching the local context for variables of certain expected types. We currently already do this in multiple places, and building an abstraction around it allows us to deduplicate code, making it more robust, more powerful and more performant (although the performance win is probably negligible, given we don't search the local context that often).

* I've built a datastructure that stores a set of patterns to search for, this allows us to cleanly separate the declaration of what to search for from doing the actual search.
* Using this abstraction, I've reimplemented `SymContext.fromLocalContext`, to now do only two passes over the local context, instead of a separate pass for each variable we look for.
* I've also renamed the former to `fromMainContext` to emphasize the fact we search the context of the main (TacticM) goal, not the ambient MetaM local context, and
* I've removed the definition that canonicalized types (canonicalizeHypothesisTypes), since we've incorporated that functionality in `searchLCtx` as well, as an option on search patterns

---

In future PRs, we'd like to:
- [ ] extract an `AxEffects.fromLocalContext` using this search abstraction
  - [ ] use the former in the stand-alone `explodeStep` tactic, and
- [ ] use the search abstraction in `sym_aggregate`

### Testing:

conformance testing successful on an Aarch64 machine? WIP

### License:

### DIFF
--- a/Benchmarks/Command.lean
+++ b/Benchmarks/Command.lean
@@ -7,21 +7,38 @@ import Lean
 
 open Lean Parser.Command Elab.Command
 
+initialize
+  registerOption `benchmark {
+    defValue := false
+    descr := "enables/disables benchmarking in `withBenchmark` combinator"
+  }
+
+variable {m} [Monad m] [MonadLiftT BaseIO m] in
+def withHeartbeatsAndMs (x : m α) : m (α × Nat × Nat) := do
+  let start ← IO.monoMsNow
+  let (a, heartbeats) ← withHeartbeats x
+  let endTime ← IO.monoMsNow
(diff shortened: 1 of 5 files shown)


alexkeizer:
@bollu, this is ready for review, modulo the fact that it's stacked on #187. Could you do a pass already? This one is a bit tricky, so I'd appreciate your thoughts!

alexkeizer:
Rebased on ready for review

shigoel (approved):

shigoel:
@alexkeizer Could you please hit Squash and merge if/when the CI succeeds?

PR #191, opened 2024-09-25
Refactor to use `extractLsb'` instead of `extractLsb`

pennyannn:
## Description:

This PR refactors to use `extractLsb'` instead of `extractLsb` in the whole repository. Doing so, we are able to simplify the proofs and remove unnecessary hypothesis inputs to several functions.

### Testing:

conformance testing successful on an Aarch64 machine? Yes.

### License:

### DIFF
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -282,11 +282,7 @@ abbrev ror (x : BitVec n) (r : Nat) : BitVec n :=
     the `n`-bit bitvector `x`. -/
 @[bitvec_rules]
 abbrev lsb (x : BitVec n) (i : Nat) : BitVec 1 :=
-  -- TODO: We could use
-  --   BitVec.extractLsb' i 1 x
-  -- and avoid the cast here, but unfortunately, extractLsb' isn't supported
-  -- by LeanSAT.
-  (BitVec.extractLsb' i 1 x).cast (by omega)
+  BitVec.extractLsb' i 1 x
 
 abbrev partInstall (hi lo : Nat) (val : BitVec (hi - lo + 1)) (x : BitVec n): BitVec n :=
   let mask := allOnes (hi - lo + 1)
@@ -630,19 +626,22 @@ theorem append_of_extract_general_nat (high low n vn : Nat) (h : vn < 2 ^ n) :
 
 theorem append_of_extract (n : Nat) (v : BitVec n)
   (high0 : high = n - low) (h : high + low = n) :
-  BitVec.cast h (zeroExtend high (v >>> low) ++ extractLsb' 0 low v) = v := by
+  zeroExtend high (v >>> low) ++ extractLsb' 0 low v = BitVec.cast h.symm v := by
   ext
   subst high
   have vlt := v.isLt
   have := append_of_extract_general_nat (n - low) low n (BitVec.toNat v) vlt
   have low_le : low ≤ n := by omega
-  simp_all [toNat_zeroExtend, Nat.sub_add_cancel, low_le]
+  simp only [toNat_append, toNat_truncate, toNat_ushiftRight, extractLsb'_toNat,
+    Nat.shiftRight_zero, toNat_cast]
+  -- simp_all [toNat_zeroExtend, Nat.sub_add_cancel, low_le]
   rw [Nat.mod_eq_of_lt (b := 2 ^ n)] at this
-  · rw [this]
+  · sorry -- rw [this]
   · rw [Nat.shiftRight_eq_div_pow]
     exact Nat.lt_of_le_of_lt (Nat.div_le_self _ _) vlt
   done
 
+@[bitvec_rules]
 theorem append_of_extract_general (v : BitVec n) :
   (zeroExtend high (v >>> low)) ++ extractLsb' 0 low v =
   extractLsb' 0 (high + low) v := by
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -539,6 +539,12 @@ theorem zeroExtend_if_false [Decidable p] (x : BitVec n)
   (zeroExtend (if p then a else b) x) = BitVec.cast h_eq (zeroExtend b x) := by
   simp only [toNat_eq, toNat_truncate, ← h_eq, toNat_cast]
 
+theorem extractLsb_eq (x : BitVec n) (h : n = n - 1 + 1) :
+  BitVec.extractLsb (n - 1) 0 x = BitVec.cast h x := by
+  unfold extractLsb extractLsb'
+  ext1
+  simp [←h]
+
 theorem extractLsb'_eq (x : BitVec n) :
   BitVec.extractLsb' 0 n x = x := by
   unfold extractLsb'
@@ -626,17 +632,15 @@ theorem append_of_extract_general_nat (high low n vn : Nat) (h : vn < 2 ^ n) :
 
 theorem append_of_extract (n : Nat) (v : BitVec n)
   (high0 : high = n - low) (h : high + low = n) :
-  zeroExtend high (v >>> low) ++ extractLsb' 0 low v = BitVec.cast h.symm v := by
+  BitVec.cast h (zeroExtend high (v >>> low) ++ extractLsb' 0 low v) = v := by
   ext
   subst high
   have vlt := v.isLt
   have := append_of_extract_general_nat (n - low) low n (BitVec.toNat v) vlt
   have low_le : low ≤ n := by omega
-  simp only [toNat_append, toNat_truncate, toNat_ushiftRight, extractLsb'_toNat,
-    Nat.shiftRight_zero, toNat_cast]
-  -- simp_all [toNat_zeroExtend, Nat.sub_add_cancel, low_le]
+  simp_all [toNat_zeroExtend, Nat.sub_add_cancel, low_le]
   rw [Nat.mod_eq_of_lt (b := 2 ^ n)] at this
-  · sorry -- rw [this]
+  · rw [this]
   · rw [Nat.shiftRight_eq_div_pow]
     exact Nat.lt_of_le_of_lt (Nat.div_le_self _ _) vlt
   done
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -933,6 +933,16 @@ theorem extractLsByte_def (val : BitVec w₁) (n : Nat) :
     val.extractLsByte n = val.extractLsb' (n * 8) 8 := rfl
 
 -- TODO: upstream
+theorem extractLsb_or (x y : BitVec w₁) (n : Nat) :
+    (x ||| y).extractLsb n lo = (x.extractLsb n lo ||| y.extractLsb n lo) := by
+  apply BitVec.eq_of_getLsbD_eq
+  simp only [getLsbD_extract, getLsbD_or]
+  intros i
+  by_cases h : (i : Nat) ≤ n - lo
+  · simp only [h, decide_True, Bool.true_and]
+  · simp only [h, decide_False, Bool.false_and, Bool.or_self]
+
+-- TODO: upstream
 theorem extractLsb'_or (x y : BitVec w₁) (n : Nat) :
     (x ||| y).extractLsb' lo n = (x.extractLsb' lo n ||| y.extractLsb' lo n) := by
   apply BitVec.eq_of_getLsbD_eq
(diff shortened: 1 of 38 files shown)


Arm/BitVec.lean
```diff
@@ -289,7 +286,7 @@ abbrev lsb (x : BitVec n) (i : Nat) : BitVec 1 :=
   --   BitVec.extractLsb' i 1 x
   -- and avoid the cast here, but unfortunately, extractLsb' isn't supported
   -- by LeanSAT.
-  (BitVec.extractLsb i i x).cast (by omega)
+  (BitVec.extractLsb' i 1 x).cast (by omega)
```
shigoel, line 289:
```suggestion
  BitVec.extractLsb' i 1 x
```
No need of the `cast`

shigoel: Also, remove the outdated comment from L285-288 above.

Arm/BitVec.lean
```diff
@@ -636,12 +629,11 @@ theorem append_of_extract_general_nat (high low n vn : Nat) (h : vn < 2 ^ n) :
   done
 
 theorem append_of_extract (n : Nat) (v : BitVec n)
-  (high0 : high = n - low) (low0 : 1 <= low)
-  (h : high + (low - 1 - 0 + 1) = n) :
-  BitVec.cast h (zeroExtend high (v >>> low) ++ extractLsb (low - 1) 0 v) = v := by
+  (high0 : high = n - low) (h : high + low = n) :
+  BitVec.cast h (zeroExtend high (v >>> low) ++ extractLsb' 0 low v) = v := by
```
shigoel, line 633: This reminds me: this'd probably be a better rule if we put the cast on the RHS. Mind doing that please, @pennyannn? :)

Arm/BitVec.lean
```diff
@@ -651,17 +643,13 @@ theorem append_of_extract (n : Nat) (v : BitVec n)
     exact Nat.lt_of_le_of_lt (Nat.div_le_self _ _) vlt
   done
 
-theorem append_of_extract_general (v : BitVec n)
-  (low0 : 1 <= low)
-  (h1 : high = width)
-  (h2 : (high + low - 1 - 0 + 1) = (width + (low - 1 - 0 + 1))) :
-  BitVec.cast h1 (zeroExtend high (v >>> low)) ++ extractLsb (low - 1) 0 v =
-  BitVec.cast h2 (extractLsb (high + low - 1) 0 v) := by
+theorem append_of_extract_general (v : BitVec n) :
```
shigoel, line 646: Maybe a `bitvec_rules` tag here?

Arm/BitVec.lean
pennyannn: Done.

Arm/BitVec.lean
pennyannn: Hi @shigoel , I tried this and it broke the proof for `read_mem_bytes_of_write_mem_bytes_same` which has a goal that looks like:
```
n✝ n : Nat
     hn : Nat.succ 0 ≤ n
     n_ih : ∀ {addr : BitVec 64} {v : BitVec (n * 8)} {s : ArmState},
       n ≤ 2 ^ 64 → ¬n = 0 → read_mem_bytes n addr (write_mem_bytes n addr v s) = v
     addr : BitVec 64
     v : BitVec ((n + 1) * 8)
     s : ArmState
     hn1 : n + 1 ≤ 2 ^ 64
     hn0 : ¬n + 1 = 0
     ⊢ BitVec.cast ⋯ (zeroExtend (n * 8) (v >>> 8) ++ read_mem addr (write_mem addr (extractLsb' 0 8 v) s)) = v
```
Notice how the `BitVec.cast` is on the LHS. 
Could you shed some light for why putting `BitVec.cast` on RHS would be better?

shigoel: Ah, sorry! I was going by the rule of thumb that the LHS should be general enough to match and maybe the cast was limiting it. In this case, it turns out this rule was aiding a very specific theorem, so maybe the change I suggested was not great.

Sorry for the noise then!

shigoel:
A general comment @pennyannn: while I'm completely onboard with using `extractLsb'` in _definitions_ in LNSym, we should not remove _theorems_ we already proved about `extractLsb` (e.g., `extractLsb'_eq` replaces `extractLsb_eq`).  Let's keep both theorems around, in case we ever run into a case where we need to use `extractLsb`.

What say?

Arm/BitVec.lean
pennyannn: It seems like it is only used that theorem in memory proofs. Maybe let's see if there will be more use later.

Arm/BitVec.lean
```diff
@@ -59,10 +59,7 @@ attribute [bitvec_rules] BitVec.getLsbD_truncate
 attribute [bitvec_rules] BitVec.zeroExtend_zeroExtend_of_le
 attribute [bitvec_rules] BitVec.truncate_truncate_of_le
 attribute [bitvec_rules] BitVec.truncate_cast
-attribute [bitvec_rules] BitVec.extractLsb_ofFin
-attribute [bitvec_rules] BitVec.extractLsb_ofNat
 attribute [bitvec_rules] BitVec.extractLsb'_toNat
-attribute [bitvec_rules] BitVec.extractLsb_toNat
 attribute [bitvec_rules] BitVec.getLsbD_extract
 attribute [bitvec_rules] BitVec.toNat_allOnes
```
alexkeizer, line 64: Let's not get rid of these simp-lemmas, just in case `extractLsb` still shows up somewhere

Arm/BitVec.lean
```diff
@@ -951,16 +945,31 @@ theorem extractLsb_or (x y : BitVec w₁) (n : Nat) :
...
+theorem extractLsb'_or (x y : BitVec w₁) (n : Nat) :
+    (x ||| y).extractLsb' lo n = (x.extractLsb' lo n ||| y.extractLsb' lo n) := by
+  apply BitVec.eq_of_getLsbD_eq
+  simp only [getLsbD_extract, getLsbD_or]
+  intros i
+  simp only [getLsbD_extractLsb', Fin.is_lt, decide_True, getLsbD_or, Bool.true_and]
+
+-- TODO: upstream
+protected theorem extractLsb'_ofNat (x n : Nat) (l lo : Nat) :
+  extractLsb' lo l (BitVec.ofNat n x) = .ofNat l ((x % 2^n) >>> lo) := by
```
alexkeizer, line 958:
```suggestion
    extractLsb' lo l (BitVec.ofNat n x) = .ofNat l ((x % 2^n) >>> lo) := by
```

Arm/Insts/DPSFP/Advanced_simd_three_different.lean
```diff
@@ -37,8 +37,8 @@ def pmull_op (e : Nat) (esize : Nat) (elements : Nat) (x : BitVec n)
   if h₀ : elements <= e then
     result
   else
-    let element1 := elem_get x e esize H
-    let element2 := elem_get y e esize H
+    let element1 := elem_get x e esize
+    let element2 := elem_get y e esize
```
alexkeizer, line 41: Can we drop the `H` parameter altogether here? It only seems to be explicitly used in the recursive call to `pmull_op`. It might be picked up by one of the `omega` calls as well. Still, worth a try

bollu:
Nice, I like quite a bit just how clean it looks now, especially because computing with `length` is so much more intuitive.

Proofs/Bit_twiddling.lean
```diff
@@ -268,7 +268,7 @@ def parity32_impl (x : BitVec 32) : BitVec 32 :=
   (0x00006996#32 >>> x4) &&& 1#32
 
 theorem parity32_correct (x : BitVec 32) :
-  (parity32_spec x) = ((parity32_impl x).getLsb 0) := by
+  (parity32_spec x) = ((parity32_impl x).getLsbD 0) := by
```
alexkeizer, line 271:
```suggestion
  (parity32_spec x) = ((parity32_impl x)[0]) := by
```
If the index is statically known to be in-bounds, using GetElem notation might allow for more rewrites

Arm/Memory/MemoryProofs.lean
```diff
@@ -886,65 +862,45 @@ theorem entire_memory_subset_legal_regions_eq_addr
...
-         BitVec.toNat (addr2 - addr2) * 8 + 1
-        =
-       x * 8) :
+private theorem read_mem_bytes_of_write_mem_bytes_subset_n2_eq_alt_helper (val : BitVec (x * 8)):
   val =
-    BitVec.cast h
-      (extractLsb ((BitVec.toNat (addr2 - addr2) + x) * 8 - 1)
-        (BitVec.toNat (addr2 - addr2) * 8) val) := by
+    extractLsb' ((BitVec.toNat (addr2 - addr2)) * 8) (x * 8) val := by
   ext
```
alexkeizer, line 868: This theorem feels like it shouldn't be needed anymore:
- `addr2 - addr2` should simplify to `0`, via `BitVec.sub_self`
- `(BitVec.toNat 0) * 8` should constant fold to just `0`
- The final bit I'm not sure we have a simp-lemma for already, but `extractLsb' 0 w val`, where `val : BitVec w` should simplify to just `val`

Thus, if we add the last, simpler, simp-lemma, that should subsume this one

Specs/AESArm.lean
```diff
@@ -110,21 +110,20 @@ def SubWord (w : BitVec WordSize) : BitVec WordSize :=
...
+  match i with
+  | 0 => acc
+  | i' + 1 =>
+    let i := Param.Nk - i
+    let wd := extractLsb' (i * 32) 32 key
     let (x:KeySchedule) := [wd]
-    have _ : Param.Nk - (i + 1) < Param.Nk - i := by omega
-    AESArm.InitKey (Param := Param) (i + 1) key (acc ++ x)
-  termination_by (Param.Nk - i)
+    AESArm.InitKey (Param := Param) i' key (acc ++ x)
```
alexkeizer, line 119: I'm having a bit of trouble following what this code does exactly, but I assume you've checked the new version computes the same value as the old version?

alexkeizer (commented):
LGTM, minus some minor nits

Arm/Insts/DPSFP/Advanced_simd_three_different.lean
pennyannn: Yes, it is picked up by one of the omega call (in `h_2`), which in turn is used in `elem_set`.

Proofs/Bit_twiddling.lean
pennyannn: Got the following counter-example when switching to using `getElem`:
```
The prover found a potential counterexample, consider the following assignment:
  x = 0xffffffff#32
  ofBool
    (27030#32 >>>
          ((x ^^^ x >>> 16 ^^^ (x ^^^ x >>> 16) >>> 8 ^^^ (x ^^^ x >>> 16 ^^^ (x ^^^ x >>> 16) >>> 8) >>> 4) &&&
            15#32) &&&
        1#32)[0] = 0x1#1
```
Is this unexpected?

Arm/Memory/MemoryProofs.lean
pennyannn: You are right! The new proof is:
```
simp only [BitVec.sub_self, toNat_ofNat, Nat.reducePow, Nat.zero_mod, Nat.zero_mul]
exact Eq.symm (extractLsb'_eq val)
```

Specs/AESArm.lean
pennyannn: This function divide `key` (For example for AES-128, `key` is of length 128 bit) into a list of 32-bits (called seeded `KeySchedule`). The original code goes from beginning to the end of the list, and when rewriting to use structural recursion, it became a bit weird. This is because structural recursion goes from larger `i` to smaller `i`.

Check the new commit. I rewrote this function to go from end of the list to the beginning, which simplified the code. 

The same thing happens in `KeyExpansion_helper`, but that function is a recurrence function where later values in the list depends on previously generated values. So, I couldn't just change it to go from end to the beginning.

Proofs/Bit_twiddling.lean
alexkeizer: That is indeed unexpected. Let's stick with getLsbD, then.

Is the goalstate easily minimized? It'd be good to figure out why this happens regardless, so an MWE would be nice

shigoel (approved):

Proofs/Bit_twiddling.lean
pennyannn: Yes. The following would fail
```
theorem crock (x : BitVec 32) :
  x.getLsbD 0 = x[0] := by bv_decide
```
with the error
```
The prover found a potential counterexample, consider the following assignment:
  x = 0xffffffff#32
  ofBool x[0] = 0x0#1
```

Issue #162, opened 2024-09-18
provenance tracking for memory regions


bollu:
As requested by @pennyannn , we want to be able to ensure that reads in our memory model only occur from pre-defined read regions. This automatically ensures that the program has not modified memory locations that are outside of the pre-defined readable regions.

PR #192, opened 2024-09-25
feat: MemoryTree: A new data structure for improving simp_mem's complexity

bollu:
We add a data structure, `MemoryTree`, that keeps track of the current information we have about memory spans. This should allow us to more efficiently exploit e.g. `PairwiseSeparate` proofs, by lazily unfolding the `PairwiseSeparate` proof as necessary, instead of the current way, where we explode the full state.


shigoel:
I like where this may be going. For now, all I'll say is that remember this issue in case that affects design decisions:
https://github.com/leanprover/LNSym/issues/162

PR #193, opened 2024-09-26
feat: bitvector constant folding simprocs

bollu:
### Description:

This simproc simplifies expressions by attempting to constant fold using `omega` to make sure we never see modulos, and moves all arithmetic constants to the left.

The goal is for this to be upstreamed sooner rather than later.

### Testing:

Conformance passes locally. 

### License:


Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+-/
+
+/-
+This file implements bitvector expression simplification simprocs.
+We perform the following additional changes:
+
+1. Canonicalizing bitvector expression to always have constants on the left.
+  Recall that the default associativity of addition is to the left: x + y + z = (x + y) + z.
+  If we thus normalize our expressions to have constants on the left,
+  and if we constFoldiate our additions to be to the left, we will naturally perform
```
alexkeizer, line 14:
```suggestion
  and if we constant-fold our additions to be to the left, we will naturally perform
```

alexkeizer: This sentence still feels a bit broken, but either way, please rephrase constFoldiate

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+  constant folding:
+
+  (a) (x + c) -> (c + x).
+  (b) x + (y + z) -> (x + y) + c.
```
alexkeizer, line 18:
```suggestion
  (b) x + (y + z) -> (x + y) + z.
```

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+private def mkSubNat (x y : Expr) : Expr :=
+  let lz := levelZero
+  let nat := mkConst ``Nat
+  let instSub := mkConst ``instSubNat
+  let instHSub := mkAppN (mkConst ``instHSub [lz]) #[nat, instSub]
+  mkAppN (mkConst ``HSub.hSub [lz, lz, lz]) #[nat, nat, nat, instHSub, x, y]
+
+/--
+Given an expression of the form `n#w`, return the value of `n` if it is a ground constant.
+-/
```
alexkeizer, line 59:
```suggestion
/--
Given an expression of the form `n#w`, return the value of `n` if it is a ground constant.

Notice that this is different from `getBitVecValue?` in that here we allow `w` to be symbolic.
Hence, we might not know the width, explaining why we return a `Nat` rather than a `BitVec`.
-/
```

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+@[inline] def proveByOmega (ty : Expr) : SimpM Step := do
+  let proof : Expr ← mkFreshExprMVar ty
+  let g := proof.mvarId!
+  let some g ← g.falseOrByContra
+    | return .continue
+  try
+    g.withContext (do Lean.Elab.Tactic.Omega.omega (← getLocalHyps).toList g {})
+  catch _ =>
+    return .continue
+  return .done { expr := ty, proof? := proof }
```
alexkeizer, line 87: Is it just me, or are these two definition exactly the same? I don't see where the first invokes `bv_omega`.

I'm also not sure about the re-use of `Step` here. That type has specific meaning for simplification and expression traversal, which doesn't seem appropriate for a definition that purports to being a general way to prove things by omega. I'd be more satisified if we call it `dischargeByOmega` and explicate in the docstring that this is specifically intended to discharge side-conditions in simp-procs.

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+
+def processingEmoji : String := "⚙️"
```
alexkeizer, line 89: Should this go to some Common file?

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+
+-- x % n = x if x < n
+@[inline] def reduceModOfLt (x : Expr) (n : Expr) : SimpM Step := do
+  trace[Tactic.address_normalization] "{processingEmoji} reduceModOfLt '{x} % {n}'"
+  let ltTy := mkLTNat x n
+  let Step.done { expr := _, proof? := some p} ← proveByOmega ltTy
+    | return .continue
+  let eqProof ← mkAppM ``Nat.mod_eq_of_lt #[p]
+  trace[Tactic.address_normalization] "{checkEmoji} reduceModOfLt '{x} % {n}'"
+  return .done { expr := x, proof? := eqProof : Result }
```
alexkeizer, line 99: I'm really looking forward to a version of simp that can take auto-tactics as dischargers, that would remove the need for at least a few of these simprocs.

In the short term, yes, let's add them to LNSym, but for the upstreaming story, could we develop something slightly more general which takes in a simp-lemma with auto-tactics, and creates a simproc that does the right thing?

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+theorem BitVec.add_ofNat_eq_ofNat_add (n : Nat) (x : BitVec w) : x + BitVec.ofNat w n = BitVec.ofNat w n + x := by
+  apply BitVec.add_comm
+
+simproc [address_normalization] moveAddConstLeft ((_ + _ : BitVec _)) :=
+  moveBinConstLeft ``HAdd.hAdd 6 ``BitVec.add_comm ``BitVec.add_ofNat_eq_ofNat_add
+
+@[address_normalization]
+theorem BitVec.ofNat_add_ofNat_eq_add_ofNat (w : Nat) (n m : Nat) : BitVec.ofNat w n + BitVec.ofNat w m = BitVec.ofNat w (n + m) := by
+  apply BitVec.eq_of_toNat_eq
+  simp
```
alexkeizer, line 178: We only want to do this if `n` and `m` are constants, right?

I guess it's pretty uncommon to have `x#w + 10#w` for `x` an fvar, but we should still consider it a possibility

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+
+  -- simp made no progress
+
+/-- Reassociate addition to left. -/
+@[address_normalization]
+theorem BitVec.constFold_add (x y z : BitVec w) : x + (y + z) = x + y + z := by
+  rw [BitVec.add_assoc]
```
alexkeizer, line 185: That's a deceptive name: it's not constant folding, right, just re-associating?
```suggestion
@[address_normalization]
theorem BitVec.add_assoc_symm (x y z : BitVec w) : x + (y + z) = x + y + z := by
  rw [BitVec.add_assoc]
```

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+  simp only [address_normalization]
+
+/-- info: 'eg₇' depends on axioms: [propext] -/
+#guard_msgs in #print axioms eg₇
+
+theorem eg₈ : 100#w + 200#w = 300#w := by
+  simp only [address_normalization]
+
+/-- info: 'eg₈' depends on axioms: [propext] -/
+#guard_msgs in #print axioms eg₈
```
alexkeizer, line 267: I'd prefer we move this to a file under the `Tests` directory

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+
+simproc [address_normalization] moveAddConstLeft ((_ + _ : BitVec _)) :=
+  moveBinConstLeft ``HAdd.hAdd 6 ``BitVec.add_comm ``BitVec.add_ofNat_eq_ofNat_add
+
+@[address_normalization]
+theorem BitVec.ofNat_add_ofNat_eq_add_ofNat (w : Nat) (n m : Nat) : BitVec.ofNat w n + BitVec.ofNat w m = BitVec.ofNat w (n + m) := by
+  apply BitVec.eq_of_toNat_eq
+  simp
+
+  -- simp made no progress
```
alexkeizer, line 180:
```suggestion
```
Stray comment

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+    if let .done res ← reduceModSub x n then
+      return .done res
+    return .continue
+  | _ => do
+     return .continue
+
+simproc↑ [address_normalization] reduce_mod_omega (_ % _) := fun e => reduceMod e
+
+/-- In a commutative binary operation, move a bitvector constant to the left. -/
+@[inline, bv_toNat] def moveBinConstLeft (declName : Name) -- operator to constant fold, such as `HAdd.hAdd`.
```
alexkeizer, line 132: It's not just moving, it's also constant folding, right?

Let's call it `canonicalizeBinConst`, and explicitly mention in the docstring that this:
- will constant fold if both operands are constant, or
- move the constant left, if there is exactly one

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,267 @@
...
+Authors: Siddharth Bhat, Tobias Grosser
+-/
+
+/-
+This file implements bitvector expression simplification simprocs.
+We perform the following additional changes:
+
+1. Canonicalizing bitvector expression to always have constants on the left.
+  Recall that the default associativity of addition is to the left: x + y + z = (x + y) + z.
+  If we thus normalize our expressions to have constants on the left,
```
alexkeizer, line 13: We actually only canonicalize addition, right, not multiplication yet? We should mention that somewhere, either here, or as a TODO further down

alexkeizer (changes requested):
Your PR description confused/scared me a bit, because of the bit where you say "by attempting to constant fold using omega". You should maybe write a bit more clearly that (a) we simplify modulo operations `a % m` when `omega` can prove that `a` is  smaller than `m`, and (b) we constant fold.

Or did I misunderstand the code and is `omega` actually used for the constant folding part also?

Arm/Memory/AddressNormalization.lean
bollu: @alexkeizer I think we would still want `(x + 10)#w`, since the reason we have `x : Nat` is (ostensibly) to have it be treated like a `Nat`.

Arm/Memory/AddressNormalization.lean
bollu: Done!

Arm/Memory/AddressNormalization.lean
bollu: Cleaned this up, renamed to `dischargedByOmega` and added a docstring.

Arm/Memory/AddressNormalization.lean
bollu: I thought about this, maybe it should live in `Tactics/Common`, so I moved it there.

Arm/Memory/AddressNormalization.lean
bollu: @alexkeizer that makes a ton of sense.

bollu:
@alexkeizer all comments addressed!

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,204 @@
...
+@[inline] def dischargeByOmega (ty : Expr) : SimpM Step := do
+  let proof : Expr ← mkFreshExprMVar ty
+  let g := proof.mvarId!
+  let some g ← g.falseOrByContra
+    | return .continue
+  try
+    g.withContext (do Lean.Elab.Tactic.Omega.omega (← getLocalHyps).toList g {})
+  catch _ =>
+    return .continue
+  return .done { expr := ty, proof? := proof }
```
alexkeizer, line 84:
```suggestion
@[inline] def dischargeByOmega (ty : Expr) : SimpM (Option Expr) := do
  let proof : Expr ← mkFreshExprMVar ty
  let g := proof.mvarId!
  let some g ← g.falseOrByContra
    | return none
  try
    g.withContext (do Lean.Elab.Tactic.Omega.omega (← getLocalHyps).toList g {})
  catch _ =>
    return none
  return some proof
```

On second thought, I still really dislike the use of `Step` here. Especially since you match specifically on `done`, but the usual semantics of `Step` would also allow for a return value with `visit`. 

I understand the desire for a more semantically meaningful type than `Option`, but `Step` has the *wrong* semantics, and I'd thus consider it actively worse than `Option`.

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,204 @@
...
+
+-- x % n = x if x < n
+@[inline] def reduceModOfLt (x : Expr) (n : Expr) : SimpM Step := do
+  trace[Tactic.address_normalization] "{processingEmoji} reduceModOfLt '{x} % {n}'"
+  let ltTy := mkLTNat x n
+  let Step.done { expr := _, proof? := some p} ← dischargeByOmega ltTy
+    | return .continue
```
alexkeizer, line 91:
```suggestion
  let some p ← dischargeByOmega ltTy
    | return .continue
```

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,204 @@
...
+  let eqProof ← mkAppM ``Nat.mod_eq_of_lt #[p]
+  trace[Tactic.address_normalization] "{checkEmoji} reduceModOfLt '{x} % {n}'"
+  return .done { expr := x, proof? := eqProof : Result }
+
+-- x % n = x - n if x >= n and x - n < n
+@[inline] def reduceModSub (x : Expr) (n : Expr) : SimpM Step := do
+  trace[Tactic.address_normalization] "{processingEmoji} reduceModSub '{x} % {n}'"
+  let geTy := mkGENat x n
+  let Step.done { expr := _, proof? := some geProof} ← dischargeByOmega geTy
+    | return .continue
```
alexkeizer, line 101:
```suggestion
  let some geProof ← dischargeByOmega geTy
    | return .continue
```

Arm/Memory/AddressNormalization.lean
```diff
@@ -0,0 +1,204 @@
...
+  let subTy := mkSubNat x n
+  let ltTy := mkLTNat subTy n
+  let Step.done { expr := _, proof? := some ltProof} ← dischargeByOmega ltTy
+    | return .continue
```
alexkeizer, line 105:
```suggestion
  let some ltProof ← dischargeByOmega ltTy
    | return .continue
```

shigoel:
@bollu Could you please squash and merge if CI succeeds?

PR #195, opened 2024-09-26
feat: replace use of (evalTactic bv_omega) with a manual run of simp followed by (evalTactic omega)

bollu:
### Description:

We're attempting to make `simp_mem` faster by controlling the evaluation of both `simp` and `omega`. Toward this, as a first step, we invoke `simp` ourselves.

### Testing:

Conformance succeeds. 

### License:


Arm/Memory/SeparateAutomation.lean
```diff
@@ -121,9 +121,18 @@ structure SimpMemConfig where
 structure Context where
   /-- User configurable options for `simp_mem`. -/
   cfg : SimpMemConfig
-
-def Context.init (cfg : SimpMemConfig) : Context where
-  cfg := cfg
+  /-- Cache of `bv_toNat` simp context. -/
+  bvToNatSimpCtx : Simp.Context
+  /-- Cache of `bv_toNat` simprocs. -/
+  bvToNatSimprocs : Array Simp.Simprocs
```
alexkeizer, line 127: Should we bundle these in a single struct, given that we basically always have these two types together? (Why are simprocs not just part of `Simp.Context` in the first place?)

Tactics/Simp.lean
```diff
@@ -52,12 +52,15 @@ def LNSymSimpContext
...
+  -- Whether the default simprocs should be used.
+  (useDefaultSimprocs : Bool := true)
   -- argument to `DiscrTree.mkPath`
   (noIndexAtArgs : Bool := true)
   : MetaM (Simp.Context ×  Array Simp.Simprocs) := do
   let mut ext_simpTheorems := #[]
   let default_simprocs ← Simp.getSimprocs
-  let mut all_simprocs := (#[default_simprocs] : Simp.SimprocsArray)
+  let mut all_simprocs :=
+    if useDefaultSimprocs then #[default_simprocs] else #[]
```
alexkeizer, line 63: Why not something like:
```
let mut all_simprocs := #[]
if useDefaultSimprocs then
  all_simprocs.push <|<- Simp.getSimprocs
```
So we don't needlessly fetch the simprocs if not needed

```

File: Tactics/Simp.lean [hunk 3]
```diff
@@ -113,4 +116,21 @@ def LNSymSimp (goal : MVarId)
...
+    : MetaM (Option MVarId) := do
+   g.withContext do
+    let fvars : Array FVarId :=
+      (← getLCtx).foldl (init := #[]) fun fvars d => fvars.push d.fvarId
+    let (result, _stats) ← simpGoal g ctx simprocs (fvarIdsToSimp := fvars)
+      (simplifyTarget := true) (discharge? := none)
+    match result with
+    | none => return none
+    | some (_newHyps, g') => pure g'
+
```
alexkeizer, line 135: Should we add an argument similar to [`Tactic.Location`](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/Tactic/Location.html#Lean.Elab.Tactic.Location) to LNSymSimp instead of creating a new single-purpose def?

Location probably doesn't work as-is, because it takes explicitly provided local variables as `Syntax` for some reason, but we can copy the shape of the type

bollu: @alexkeizer I'm not sure, to be honest, I prefer this separation, because it makes it clear that `LNSymSimpAtStar` is a think wrapper around the existing `LNSymSimp`.

Arm/Memory/SeparateAutomation.lean
bollu: @alexkeizer I'm not sure why it's not part of `Simp.Context`, and that makes a little hesitant to bundle them together :)

Tactics/Simp.lean
bollu: We get a weird failure here:

```lean
def fails : MetaM Unit := do
  -- arr : Array arr✝
  let mut arr : Array Bool ← do
      pure #[]
  -- unknown identifier arr
  arr := arr.push false
  return ()

def suceeds : MetaM Unit := do
  let arr : Array Bool ← do
      pure #[]
  let mut arr := arr
  arr := arr.push false
  return ()
```

bollu: @alexkeizer  I can do:

```
  let all_simprocs : Array Simp.Simprocs ← do
    if useDefaultSimprocs then
      pure #[(← Simp.getSimprocs)]
    else
      pure #[]
  /- For whatever reason, can't have the `all_simprocs` be `mut`. -/
  let mut all_simprocs := all_simprocs
```

if that's better in your view.

bollu: Addressed,

bollu:
@alexkeizer I addressed one of your comments, and I ignored two of them, because it's not clear to me that it's the right design. I prefer YAGNI, so I'm happy with the patch as-is. What do you feel?

Arm/Memory/SeparateAutomation.lean
```diff
@@ -427,10 +444,18 @@ def simpAndIntroDef (name : String) (hdefVal : Expr) : SimpMemM FVarId  := do
 
 /-- SimpMemM's omega invoker -/
 def omega : SimpMemM Unit := do
-  -- https://leanprover.zulipchat.com/#narrow/stream/326056-ICERM22-after-party/topic/Regression.20tests/near/290131280
-  -- @bollu: TODO: understand what precisely we are recovering from.
-  withoutRecover do
-    evalTactic (← `(tactic| bv_omega))
+  SimpMemM.withMainContext do
+    -- https://leanprover.zulipchat.com/#narrow/stream/326056-ICERM22-after-party/topic/Regression.20tests/near/290131280
+    -- @bollu: TODO: understand what precisely we are recovering from.
```
alexkeizer, line 449: You might want to move this comment down to the `withoutRecover` that it's talking about

Tactics/Simp.lean
```diff
@@ -52,12 +52,18 @@ def LNSymSimpContext
...
   : MetaM (Simp.Context ×  Array Simp.Simprocs) := do
   let mut ext_simpTheorems := #[]
-  let default_simprocs ← Simp.getSimprocs
-  let mut all_simprocs := (#[default_simprocs] : Simp.SimprocsArray)
+  /- Workaround for https://github.com/leanprover/lean4/issues/5607: Elaboration failure with let mut whose RHS is a do notation -/
+  let all_simprocs ← do
+    if useDefaultSimprocs then
+      pure #[← Simp.getSimprocs]
+    else pure #[]
+  let mut all_simprocs := all_simprocs
```
alexkeizer, line 66: This also works (although the use of `:=` and `←` looks a bit wonky). I'll let you pick which one you find cleaner
```suggestion
  let mut all_simprocs :=← do
    if useDefaultSimprocs then
      pure #[← Simp.getSimprocs]
    else pure #[]
```

alexkeizer (commented):
LGTM

Tactics/Simp.lean
bollu: I'll pick the current one.

shigoel (approved):

PR #196, opened 2024-09-26
chore: fix geomean computation

bollu:
### Description:

The `geomean` computation that was incorrectly suggested by @bollu failed to take into account that the `TotalRunTime` counter was adding the real runtimes, not multiplying them as one should for geomean.

We change the `geomean` computation algorithm, to accumulate the `log` of the runtimes, and then divide by `n` followed by exponentiation as a final step for better numerical accuracy.

$$
g \equiv (t_1 \times t_2 \times \dots \times t_n)^{1/n} = \exp \bigg(\frac{\log t_1 + \log t_2 + \dots \log t_n}{n} \bigg)
$$

### Testing:

Conformance succeeds.

### License:


shigoel (approved):
Thanks!

PR #198, opened 2024-09-26
feat: add combinators for more fine-grained benchmarking

alexkeizer:
### Description:

This adds combinators to benchmark specific parts of a tactic, either by directly outputting the data, or by storing datapoints in a state for aggregation. Either behaviour is controlled by a `benchmark` option, which defaults to false, but is set to `true` in the benchmark command.

### Testing:

### License:


Benchmarks/Command.lean
```diff
@@ -33,22 +50,144 @@ elab "benchmark" id:ident declSig:optDeclSig val:declVal : command => do
...
 /-- The default `maxHeartbeats` setting.
 
 NOTE: even if the actual default value changes at some point in the future,
 this value should *NOT* be updated, to ensure the percentages we've reported
 in previous versions remain comparable. -/
-def defaultMaxHeartbeats : Nat := 200000
+private def defaultMaxHeartbeats : Nat := 200000
+
+private def percentOfDefaultMaxHeartbeats (heartbeats : Nat) : Nat :=
+  heartbeats / (defaultMaxHeartbeats * 10)
```
bollu, line 63: I must admit, I don't understand the maths here: Shouldn't it return a `Float`, and be `100.0 * (heartbeats / defaultMaxHeartbeats)`

Benchmarks/Command.lean
```diff
@@ -33,22 +50,144 @@ elab "benchmark" id:ident declSig:optDeclSig val:declVal : command => do
...
+We define `withAggregatedBenchmark`, which functions like `withBenchmark`,
+except it will store a running average of the statistics in a `BenchmarkState`
+which will be reported in one go when `reportAggregatedBenchmarks` is called.
+-/
+section
+
+structure BenchmarkState.Stats where
+  totalHeartbeats : Nat := 0
+  totalTimeInMs : Nat := 0
+  samples : Nat := 0
```
bollu, line 125: With your permission, I'd like to refactor this to store the data as follows:

```lean
structure Sample where
  timeInMs : Nat
  heartbeats : Nat


structure BenchmarkState.State where
  samples : Array BenchmarkState.Samples
```

bollu:
@alexkeizer I like the idea, and I'm glad I didn't start writing the same thing before looking at the merge queue :) I would like your permission to push to the PR, to maintain the statistics in a more fine-grained fashion.

alexkeizer:
The idea, for me, is to figure out out of the $X seconds spent by `sym_n`, how many are spent on a particular phase (combined across all steps), that's why I'm basically only interested in the totals.

Keeping the individual samples is more versatile for other purposes, you're right about that. Would it make sense to track both? So keeping the totals still. Or is summing ~500 Nats fast enough that it's not really worth the complexity? Not sure what the tradeoffs there, and not sure whether they really matter.

Which is to say, feel free to refactor, so long as you give me an easy way to report the totals I'm happy

Benchmarks/Command.lean
alexkeizer: Ah, you'd think so, but there is a difference of a factor 1000 in the heartbeats as measured by Lean, and the heartbeats as reported to users. So, the math here is `(heartbeats / (defaultMaxHeartbeats * 1000) * 100`, which cancels out to what I wrote there (please do check my maths).

I guess a comment explaining this is warrented

alexkeizer: Also, why it doesn't return a float: it could, but we're not super interested in stuff that takes less than 1 percent anyway.

shigoel:
@alexkeizer @bollu Converting to draft until the comments are resolved.

Benchmarks/Command.lean
```diff
@@ -7,21 +7,38 @@ import Lean
...
 open Lean Parser.Command Elab.Command
 
+initialize
+  registerOption `benchmark {
+    defValue := false
+    descr := "enables/disables benchmarking in `withBenchmark` combinator"
+  }
+
+variable {m} [Monad m] [MonadLiftT BaseIO m] in
+def withHeartbeatsAndMs (x : m α) : m (α × Nat × Nat) := do
```
shigoel, line 17: @alexkeizer Comment here

shigoel (approved):

PR #199, opened 2024-09-26
fix: `inferStatePrefix...` should default to `s1` as the first intermediate state.

alexkeizer:
### Description:

We claim in the error that we fall back to `s1` as the first intermediate state if we can't infer anything. 
However, the implementation actually used `s2`. 

This PR fixes that by setting `currentStateNumber := 0` in the fallback case. The off-by-one behaviour is because this number is incremented *before* it's used to generate a new state.

### Testing:

### License:

### DIFF
--- a/Tactics/SymContext.lean
+++ b/Tactics/SymContext.lean
@@ -247,7 +247,7 @@ Falling back to the default numbering scheme, \
 with `s1` as the first new intermediate state"
     modifyThe SymContext ({ · with
       state_prefix := "s",
-      currentStateNumber := 1 })
+      currentStateNumber := 0 })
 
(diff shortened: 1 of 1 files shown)


bollu:
LGTM, though maybe calling it `currentStateNumber` isn't the best, how about `lastSeenStateNumber`? You're the czar of this pat of the repo, so I leave it to you.

alexkeizer:
> LGTM, though maybe calling it currentStateNumber isn't the best, how about lastSeenStateNumber? You're the czar of this pat of the repo, so I leave it to you.

Hmm, it matches the number of the `currentState` field of the `AxEffects`, so we'd have to change that one as well to be consistent, and I kind of like the `currentState` name

PR #200, opened 2024-09-26
feat: search for reads of *any* register when building an initial `SymContext`

alexkeizer:
### Description:

Stacked on #189, we use the more powerful search capabilities to include hypotheses that describe reads from *any* `StateField` into the initial AxEffects.

### Testing:

What tests have been run? Did `make all` succeed for your changes? Was
conformance testing successful on an Aarch64 machine? Yes

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #201, opened 2024-09-27
refactor: use searchLCtx in sym_aggregate

alexkeizer:
### Description:

Stacked on #200.

This uses the `searchLCtx` machinery we built in #189 in `sym_aggregate`'s implementation.
For now, this searches for exactly the same kind of expressions as before, but this makes it much easier to expand this set in the next PR.

### Testing:

### License:

### DIFF
--- a/Tactics/Aggregate.lean
+++ b/Tactics/Aggregate.lean
@@ -6,6 +6,7 @@ Author(s): Alex Keizer, Siddharth Bhat
 import Lean
 import Tactics.Common
 import Tactics.Simp
+import Tactics.Sym.LCtxSearch
 
 open Lean Meta Elab.Tactic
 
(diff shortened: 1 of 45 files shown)


Tactics/Aggregate.lean
```diff
@@ -65,44 +66,36 @@ elab "sym_aggregate" simpConfig?:(config)? loc?:(location)? : tactic => withMain
...
-              return axHyps.push decl
-            else if ← isDefEq type expectedReadMem then
-              trace[Tactic.sym] "{Lean.checkEmoji} match for {expectedReadMem}"
-              return axHyps.push decl
-            else
-              trace[Tactic.sym] "{Lean.crossEmoji} no match"
-              return axHyps
+        -- `r ?field ?state = ?rhs`
+        searchLCtxFor (whenFound := whenFound)
+          (matchUnderBinders := true)
```
shigoel, line 77: Comment -- why is this necessary, etc.?

Tactics/Aggregate.lean
```diff
@@ -65,44 +66,36 @@ elab "sym_aggregate" simpConfig?:(config)? loc?:(location)? : tactic => withMain
...
-    let expectedReadMem : MetaM Expr := do
-      let n ← mkFreshExprMVar (mkConst ``Nat)
-      let addr ← mkFreshExprMVar (mkApp (mkConst ``BitVec) (toExpr 64))
-      let mem ← mkFreshExprMVar (mkConst ``Memory)
-      let rhs ← mkFreshExprMVar none
-      mkEq (mkApp3 (mkConst ``Memory.read_bytes) n addr mem) rhs
-    let expectedAlign : MetaM Expr := do
-      let state ← mkFreshExprMVar mkArmState
-      return mkApp (mkConst ``CheckSPAlignment) state
+    let ((), axHyps) ← StateT.run (s := #[]) <|
```
shigoel, line 69: Some explanation for the nonsavvy devs would be good.

Tactics/Sym/LCtxSearch.lean
```diff
@@ -95,9 +97,14 @@ variable {m}
   def-eq to the pattern, or if `whenFound` returned `skip` for all variables
   that were found. For convenience, we pass in the `expectedType` here as well.
   See `throwNotFound` for a convenient way to throw an error here.
-- If `changeType` is set to true, then we change the type of every successful
-  match (i.e., `whenFound` returns `continu` or `done`) to be exactly the
-  `expectedType`
+- If `changeType` (default `false`) is set to true, then we change the type of
```
shigoel, line 100: @alexkeizer Please change the `SymContext` to also set `matchUnderBinders` to true to pick up on non-effects.

alexkeizer: In hindsight, this isn't as straightforward: when constructing a `SymContext` we have that the initial state and current state are the same, so a non-effect hyp would look like `forall f, f ≠ _ -> r f <state> = r f <state>`. This is, of course, a tautology (and the default thing we initialize the non-effects proof to), so there's no need to search for it.

We *could* search for `forall f, f ≠ _ -> r f <state> = r f ?newInitialState`, where the rhs refers to an mvar, and update the initial state accordingly. This gets into tricky territory, because we then have to make sure that all of the fields mentioned in the sequence of non-equalities have a value assigned to them in the other proofs. 

In conclusion, let's not tackle this particular problem in this PR. We can just keep `sym_aggregate` around in proofs that need to aggregate non-effects, and think more about this after we've solved intermediate state aggregation (since that's when having the initial state populated correctly becomes more important). For now, we can always work-around this by having explicit hypotheses `r x12 s1 = r x12 s0` hypotheses for those fields that we care about being tracked, as a workaround.

shigoel (approved):

PR #204, opened 2024-09-27
feat: add options for more control over benchmarking and better profiling support

alexkeizer:
### Description:

This PR gives more fine-grained control over benchmarrking:
- It adds an option to control how many runs we do
- We change the makefile to invoke Lean directly, instead of using Lake. This is needed to ensure we don't get cached data
- We've moved the driving of the benchmarks/profiling to a shell script (which is called from the makefile), and ensured benchmarking / profiling data gets logged to files automatically.

### Testing:

### License:


Benchmarks/Command.lean
```diff
@@ -12,58 +12,104 @@ initialize
     defValue := false
     descr := "enables/disables benchmarking in `withBenchmark` combinator"
   }
+  registerOption `benchmark.runs {
+    defValue := (5 : Nat)
+    descr := "controls how many runs the `benchmark` command does. \
+    NOTE: this value is ignored when the `profiler` option is set to true"
```
shigoel, line 18: By "this value" here, do you mean the default value of runs, or benchmarking altogether? I believe the former (and perhaps you run the benchmarks only once when the profiler is on?), but clarification would be nice.

Benchmarks/Command.lean
```diff
@@ -12,58 +12,104 @@ initialize
...
+    x
+  else
+    withTraceNode `benchmark (fun _ => pure msg) x (collapsed := false)
 
+/--
+Run a benchmark for a set number of times, and report the average runtime.
+
+If the `profiler` option is set true, we run the benchmark only once, with:
+- `trace.profiler` to true, and
+- `trace.profiler.output` set based on the `benchmark.profilerDir` and the
```
shigoel, line 49: What's `benchmark.profilerDir`?

Makefile
```diff
@@ -5,6 +5,8 @@
 SHELL := /bin/bash
 
 LAKE = lake
+LEAN = $(LAKE) env lean
+GIT = git
```
shigoel, line 9: Remove dead var -- `GIT`

scripts/benchmark.sh
shigoel, line 1: Missing copyright blurb

scripts/profile.sh
shigoel, line 1: Missing copyright blurb

lakefile.lean
```diff
@@ -33,6 +33,7 @@ lean_lib «Doc» where
   -- add library configuration options here
 
 lean_lib «Benchmarks» where
+  leanOptions := #[⟨`weak.benchmark.runs, (0 : Nat)⟩]
```
shigoel, line 36: Any documentation for what `weak` is, exactly? I found the [relevant PR here](https://github.com/leanprover/lean4/pull/4741) but I'd rather have a doc. link, if one exists and you're aware of it.

scripts/benchmark.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+LAKE=lake
+BENCH="$LAKE env lean -Dweak.benchmark.runs=5"
+OUT="data/benchmarks"
```
shigoel, line 5: Let's add an option to clean all these generated files under `clean_all` in the Makefile. Same for files generated by `profile.sh`.

scripts/benchmark.sh
shigoel: Also -- add a doc string describing what's going on here. Same for `profile.sh`

lakefile.lean
alexkeizer: I'm not aware of any docs: I got the hint to use `weak` from the error that got thrown when I tried setting `benchmark.runs` directly

Benchmarks/Command.lean
alexkeizer: Indeed, I've made the description more explicit

Benchmarks/Command.lean
alexkeizer: Oops, it's an old option that got removed. I've updated the docstring now

scripts/benchmark.sh
alexkeizer: I was a bit wary at first: I tend to run `make clean` when something breaks my build and would not want this to also delete all of my benchmarking data. I guess `make_all` is already a separate target, so it should be fine though

shigoel: @alexkeizer Indeed (but I think you meant `clean_all`, not `make_all`). Mind making that change, and then we'll hit merge.

alexkeizer: Done

alexkeizer:
@shigoel merged main and all comments addressed, this should be ready to merge

shigoel (approved):
Thanks, @alexkeizer!

PR #205, opened 2024-09-29
Syntax for stating register and memory frame conditions

shigoel:
### Description:

Introduce syntax for stating register and memory frame conditions, which specify exactly which state components are not affected by the program's execution.

The popcount proof is modified to use this new syntax.

Also: use `gen_step.print_names` option to print generated stepi theorems' names.

### Testing:

Yes

### License:

### DIFF
--- a/Arm/Syntax.lean
+++ b/Arm/Syntax.lean
@@ -17,24 +17,25 @@ macro_rules | `($s[$base,$n]) => `(read_mem_bytes $n $base $s)
 
 
 /-! Notation to specify the frame condition for non-memory state components. E.g.,
-`REGS_UNCHANGED_EXCEPT [[.GPR 0, .PC]] (sf, s0)` is sugar for
+`REGS_UNCHANGED_EXCEPT [.GPR 0, .PC] (sf, s0)` is sugar for
 `∀ f, f ∉ [.GPR 0, .PC] → r f sf = r f s0`
 -/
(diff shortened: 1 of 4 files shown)


Arm/Syntax.lean
```diff
@@ -6,11 +6,35 @@ Author(s): Siddharth Bhat
...
 namespace ArmStateNotation
 
 /-! We build a notation for `read_mem_bytes $n $base $s` as `$s[$base, $n]` -/
 @[inherit_doc read_mem_bytes]
 syntax:max term noWs "[" withoutPosition(term)  ","  withoutPosition(term) noWs "]" : term
 macro_rules | `($s[$base,$n]) => `(read_mem_bytes $n $base $s)
+
+
+/-! Notation to specify the frame condition for non-memory state components. E.g.,
+`REGS_UNCHANGED_EXCEPT [[.GPR 0, .PC]] (sf, s0)` is sugar for
```
bollu, line 20: Do we want the double paren?

```
`REGS_UNCHANGED_EXCEPT [[.GPR 0, .PC]] 
```

At least for me, I find:

```
`REGS_UNCHANGED_EXCEPT [.GPR 0, .PC] 
```

easier to read.

shigoel: Tell me how to remove the extra `[]`, and I'll do it :)

shigoel: Doing the naive thing (i.e., removing the block parens. around `$regs`) in the syntax and macro rules isn't cutting it

PR #208, opened 2024-10-01
feat: remove toNat lemmas from bitvec_rules

bollu:
This prevents us from unfolding expressions that have `.toNat` into gigantic expressions. Furthermore, this helps `bv_decide`, since it can see pure bitvector statements.

### DIFF
--- a/Arm/BitVec.lean
+++ b/Arm/BitVec.lean
@@ -43,12 +43,13 @@ attribute [bitvec_rules] BitVec.msb_zero
 attribute [bitvec_rules] BitVec.toNat_cast
 attribute [bitvec_rules] BitVec.getLsbD_cast
 attribute [bitvec_rules] BitVec.getMsbD_cast
-attribute [bitvec_rules] BitVec.toNat_ofInt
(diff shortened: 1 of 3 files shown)


Arm/BitVec.lean
```diff
@@ -43,12 +43,13 @@ attribute [bitvec_rules] BitVec.msb_zero
 attribute [bitvec_rules] BitVec.toNat_cast
 attribute [bitvec_rules] BitVec.getLsbD_cast
 attribute [bitvec_rules] BitVec.getMsbD_cast
-attribute [bitvec_rules] BitVec.toNat_ofInt
+-- attribute [bitvec_rules] BitVec.toNat_ofInt -- TODO: not tagged bv_toNat.
```
shigoel, line 46: Upstreamed the tag for this lemma?

shigoel (approved):
LGTM! Thx!

Arm/BitVec.lean
bollu: upstreamed at https://github.com/leanprover/lean4/pull/5608

PR #209, opened 2024-10-01
feat: attempt to preserve stack alignment proof in AxEffect update

alexkeizer:
### Description:

Stacked on #207.

When an write to memory or a write to a register that is not SP is made, we update the proof of stack alignment. When a write to SP is made, we create a new mvar with the proof obligation that the new value is aligned, and store that mvar in a new `sideConditions` field. 

### Testing:

### License:

### DIFF
--- a/Arm/Insts/Common.lean
+++ b/Arm/Insts/Common.lean
@@ -219,31 +219,35 @@ def CheckSPAlignment (s : ArmState) : Prop :=
 instance : Decidable (CheckSPAlignment s) := by unfold CheckSPAlignment; infer_instance
 
 @[state_simp_rules]
-theorem CheckSPAligment_w_different_eq (h : StateField.GPR 31#5 ≠ fld) :
+theorem CheckSPAlignment_w_different_eq (h : StateField.GPR 31#5 ≠ fld) :
   CheckSPAlignment (w fld v s) = CheckSPAlignment s := by
   simp_all only [CheckSPAlignment, state_simp_rules, minimal_theory, bitvec_rules]
 
-theorem CheckSPAligment_w_of_ne_sp_of (h : StateField.GPR 31#5 ≠ fld) :
+/-- A rewording of `CheckSPAlignment_w_different_eq` as an implication,
+to be used by proof automation in `AxEffects` -/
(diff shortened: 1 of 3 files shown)


Tactics/Sym/AxEffects.lean
```diff
@@ -78,6 +78,10 @@ structure AxEffects where
   However, if SP is written to, no effort is made to prove alignment of the
   new value; the field will be set to `none` -/
   stackAlignmentProof? : Option Expr
+
+  /-- `sideContitions` are proof obligations that come up during effect
```
shigoel, line 82:
```suggestion
  /-- `sideConditions` are proof obligations that come up during effect
```

shigoel: These side conditions only stem from the SP alignment proof. That deserves a mention here.

Arm/Insts/Common.lean
```diff
@@ -219,20 +219,32 @@ def CheckSPAlignment (s : ArmState) : Prop :=
 instance : Decidable (CheckSPAlignment s) := by unfold CheckSPAlignment; infer_instance
 
 @[state_simp_rules]
-theorem CheckSPAligment_of_w_different (h : StateField.GPR 31#5 ≠ fld) :
+theorem CheckSPAligment_w_different_eq (h : StateField.GPR 31#5 ≠ fld) :
   CheckSPAlignment (w fld v s) = CheckSPAlignment s := by
   simp_all only [CheckSPAlignment, state_simp_rules, minimal_theory, bitvec_rules]
 
+theorem CheckSPAligment_w_of_ne_sp_of (h : StateField.GPR 31#5 ≠ fld) :
```
shigoel, line 226: At first glance, theorems like `CheckAlignment*_sp_of` seem unnecessary to have, esp. in the face of those like `CheckSPAligment_*_eq`. Mention that they are useful to construct a non-effects proof for SP and point to `AxEffects`.

alexkeizer: I've added a comment

Tactics/Sym/AxEffects.lean
alexkeizer: I've expanded the docstring

alexkeizer:
@shigoel rebased on main and addressed comments!

PR #214, opened 2024-10-03
chore: add better trace nodes to `AxEffects`

alexkeizer:
### Description:

Stacked on: #213 

Adds tags to the `AxEffects` trace nodes, to make profile data more informative (messages don't show up in the JSON, but tags do)

### Testing:

What tests have been run? Did `make all` succeed for your changes? Was
conformance testing successful on an Aarch64 machine? Yes

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #210, opened 2024-10-02
WIP: feat: intermediate state aggregation through a persistent AxEffects object

alexkeizer:
### Description:

Stacked on
- [ ] #200 
- [ ] #207 

This PR changes `sym_n` to keep a single `AxEffects` object which is updated (rather than replaced) each step.

Currently, this significantly slows down symbolic simulation, causing timeouts when simulation more than ~60 steps of SHA512, and thus breaking a lot of proofs.

### Testing:

What tests have been run? Did `make all` succeed for your changes? Was
conformance testing successful on an Aarch64 machine?

### License:

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

PR #215, opened 2024-10-04
refactor: replace `rewrite` with `mkEqNDRec`

alexkeizer:
### Description:

Stacked on
- [ ] #214 

We've replaced uses of `rewrite` in `AxEffects` with a more manual application of `Eq.ndRec`.

This is motivated by the slowdown observed in #210, where the size of hypotheses would grow and a lot of time was being spent on rewriting. With `Eq.ndRec`, the meta-code explicitly constructs the motive, thus, no inspection of the expression being rewritten is needed, hence, the speedup.

That said, in this PR we've not yet switched to intermediate state aggregation, so the observed speedup is less significant.

TODO: put benchmark data here

### Testing:

### License:

### DIFF
--- a/Tactics/Common.lean
+++ b/Tactics/Common.lean
@@ -277,6 +277,32 @@ def Lean.Expr.eqReadField? (e : Expr) : Option (Expr × Expr × Expr) := do
     | none
   some (field, state, value)
 
+/-- Return `ArmState.program <state> = <program>` -/
+def mkEqProgram (state program : Expr) : Expr :=
+  mkApp3 (.const ``Eq [1]) (mkConst ``Program)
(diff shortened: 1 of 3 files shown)


alexkeizer:
This is currently blocked, waiting on a diagnosis of https://github.com/leanprover/lean4/issues/5610

alexkeizer:
We had a working hypothesis that the `instantiateMVars` slowdown was somehow related to proof sizes. Unfortunately, this does not seem to be the case: I compared the proof generated for 5 steps of SHA512 in this PR with the main branch, and in fact this PR generates slightly *smaller* proofs, despite being significantly slower.

PR #216, opened 2024-10-04
Simplify partInstall to use start and len

pennyannn:
### Description:

This PR simplifies the definition of `partInstall`. Its signature used to be:
```
abbrev partInstall (hi lo : Nat) (val : BitVec (hi - lo + 1)) (x : BitVec n): BitVec n
```
and now it becomes:
```
abbrev partInstall (start len : Nat) (val : BitVec len) (x : BitVec n): BitVec n
```
This follows the same pattern in `extractLsb` vs `extractLsb'`. By doing this, we are able to simplify several functions that depends on `partInstall`.

### Testing:

### License:


shigoel (approved):
Thanks, @pennyannn!

PR #219, opened 2024-10-04
WIP: refactor: reimplement initNextStep without evalTactic

alexkeizer:
### Description:

Work in progress: this currently causes a slowdown, more investigation is needed.

### Testing:

### License:

### DIFF
--- a/Arm/Exec.lean
+++ b/Arm/Exec.lean
@@ -163,6 +163,13 @@ theorem run_onestep {s s': ArmState} {n : Nat} :
   simp only [run, exists_eq_left', imp_self]
 
 /-- helper lemma for automation -/
+theorem run_of_run_succ_of_stepi_eq {s0 s1 sf : ArmState} {n : Nat}
+    (h_run : sf = run (n + 1) s0)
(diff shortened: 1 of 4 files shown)


alexkeizer:
I'm going a bit mad with this instantiateMVars nonsense. Compare a [profile](https://share.firefox.dev/4eOTB4D) from #214 with a [profile](https://share.firefox.dev/3ZUeCGK) from the current PR --- both are running the SHA512_225 benchmark.

You'll notice the former takes 8 seconds in total, of which 1.2 is spent on `init_next_step`. That was the motivation for this PR. Indeed, in the profile for this PR, `initNextStep` doesn't even make the profiler's threshold, but we can see that indeed we spend much less time executing the tactic (the `Elab.step: tacticSym_n___` node was about 3.3 seconds before, and is  1.1 seconds now).

However, all of that is lost on the metavariable instantiation. This profile was made with a toolchain version that doesn't have a node for `instantiatMVar`, but we can deduce how much is spent on it from the self time of `Elab.command`: this was 3.8 seconds, and has now blown up to 11.7 seconds.

PR #221, opened 2024-10-06
feat: generalize to arbitrary bitvec width `zeroextend_bigger_smaller` and `truncate_of_concat_is_lsb`

luisacicolini:
### Description:

This PR extends `truncate_of_concat_is_lsb_64` to arbitrary-width bitvectors. 

### Testing:

What tests have been run? Did `make all` succeed for your changes? Yes
Was conformance testing successful on an Aarch64 machine? yes 

### License:

### DIFF
--- a/Proofs/SHA512/SHA512_block_armv8_rules.lean
+++ b/Proofs/SHA512/SHA512_block_armv8_rules.lean
@@ -138,20 +138,16 @@ theorem truncate_of_concat_is_lsb_64 (x y : BitVec 64) :
   BitVec.zeroExtend 64 (x ++ y) = y := by
   bv_check "lrat_files/Sha512_block_armv8_rules.lean-truncate_of_concat_is_lsb_64-106-2.lrat"
 
-theorem truncate_of_concat_is_lsb (w : Nat) (x y : BitVec w) :
+theorem zeroExtend_append_eq (w : Nat) (x y : BitVec w) :
   BitVec.zeroExtend w (x ++ y) = y := by
   ext
-  simp [BitVec.leftshift_n_or_mod_2n]
+  simp only [truncate_append, Nat.le_refl, ↓reduceDIte, zeroExtend_eq]
 
+-- use BitVec.zeroExtend_zeroExtend_of_le for arbitrary-length bitvec
 theorem zeroextend_bigger_smaller_64 (x : BitVec 64) :
   BitVec.zeroExtend 64 (BitVec.zeroExtend 128 x) =
   BitVec.zeroExtend 64 x := by
   bv_omega
-
-theorem zeroextend_bigger_smaller (w : Nat) (x : BitVec w) :
-  BitVec.zeroExtend w (BitVec.zeroExtend (w + w) x) =
-  BitVec.zeroExtend w x := by
-  simp
 
 -- (FIXME) Generalize to arbitrary-length bitvecs.
 theorem rsh_concat_identity_128 (x : BitVec 128) :
--- a/Proofs/SHA512/SHA512_block_armv8_rules.lean
+++ b/Proofs/SHA512/SHA512_block_armv8_rules.lean
@@ -145,6 +145,11 @@ theorem zeroextend_bigger_smaller_64 (x : BitVec 64) :
   BitVec.zeroExtend 64 x := by
   bv_omega
 
+theorem zeroextend_bigger_smaller (w : Nat) (x : BitVec w) :
+  BitVec.zeroExtend w (BitVec.zeroExtend (w + w) x) =
+  BitVec.zeroExtend w x := by
+  simp
+
 -- (FIXME) Generalize to arbitrary-length bitvecs.
 theorem rsh_concat_identity_128 (x : BitVec 128) :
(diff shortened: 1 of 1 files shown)


Proofs/SHA512/SHA512_block_armv8_rules.lean
```diff
@@ -134,17 +134,25 @@ theorem concat_of_rsh_is_msb_128 (x y : BitVec 64) :
...
+  BitVec.zeroExtend w (x ++ y) = y := by
+  ext
+  simp [BitVec.leftshift_n_or_mod_2n]
+
 theorem zeroextend_bigger_smaller_64 (x : BitVec 64) :
   BitVec.zeroExtend 64 (BitVec.zeroExtend 128 x) =
   BitVec.zeroExtend 64 x := by
   bv_omega
 
+theorem zeroextend_bigger_smaller (w : Nat) (x : BitVec w) :
```
shigoel, line 151: This could be generalized further. In fact, the lemma we really need already exists in Lean --  `BitVec.zeroExtend_zeroExtend_of_le`. We should really delete `zeroextend_bigger_smaller_64` and `zeroextend_bigger_smaller`, and use that lemma instead.

Proofs/SHA512/SHA512_block_armv8_rules.lean
```diff
@@ -134,17 +134,25 @@ theorem concat_of_rsh_is_msb_128 (x y : BitVec 64) :
...
--- (FIXME) Generalize to arbitrary-length bitvecs.
 theorem truncate_of_concat_is_lsb_64 (x y : BitVec 64) :
   BitVec.zeroExtend 64 (x ++ y) = y := by
   bv_check "lrat_files/Sha512_block_armv8_rules.lean-truncate_of_concat_is_lsb_64-106-2.lrat"
 
--- (FIXME) Generalize to arbitrary-length bitvecs.
+theorem truncate_of_concat_is_lsb (w : Nat) (x y : BitVec w) :
+  BitVec.zeroExtend w (x ++ y) = y := by
+  ext
+  simp [BitVec.leftshift_n_or_mod_2n]
```
shigoel, line 144: Thanks for this! In LNSym, we prefer using `simp only` wherever possible, so could you please fix that? Note that you can also use `BitVec.truncate_append` for a smaller/cleaner proof here.

Also, I acknowledge that this file has badly named lemmas (it precedes Lean's BitVec lemma dev) --- this lemma should've really used `zeroExtend` and `append` in the name.

shigoel (commented):
Thanks for the PR, @luisacicolini!

A high-level comment: could you please use [LNSym's PR template](https://github.com/leanprover/LNSym/blob/main/.github/pull_request_template.md) for this PR's description?

luisacicolini:
thank you for reviewing this @shigoel! Everything should be fixed now. I removed `zeroextend_bigger_smaller` and added a comment pointing to the correct arbitrary-width version.

Proofs/SHA512/SHA512_block_armv8_rules.lean
```diff
@@ -134,12 +134,16 @@ theorem concat_of_rsh_is_msb_128 (x y : BitVec 64) :
...
--- (FIXME) Generalize to arbitrary-length bitvecs.
 theorem truncate_of_concat_is_lsb_64 (x y : BitVec 64) :
   BitVec.zeroExtend 64 (x ++ y) = y := by
   bv_check "lrat_files/Sha512_block_armv8_rules.lean-truncate_of_concat_is_lsb_64-106-2.lrat"
 
--- (FIXME) Generalize to arbitrary-length bitvecs.
+theorem zeroExtend_append_eq (w : Nat) (x y : BitVec w) :
+  BitVec.zeroExtend w (x ++ y) = y := by
+  ext
+  simp only [truncate_append, Nat.le_refl, ↓reduceDIte, zeroExtend_eq]
```
alexkeizer, line 144: The style in LNSym is currently a bit inconsistent, but standard Lean/Mathlib style is to indent the type/statement of a theorem with 4 spaces (vs 2 spaces for the proof itself).

Then, we can generalize the theorem further: append is a heterogenous operation and can take two bitvectors of different length! Since the value `w` and `v` is determined by the other arguments `x` and `y`, we can make the former implicit arguments.

Finally, you could consider naming this `zeroExtend_append_eq_right`, to draw extra attention to the fact that the value we're "zero-extending" to is exactly the width of the right operand, but it's also perfectly valid to argue that's the logical statement for a theorem about zeroExtend and append, and keep preferring the shorter name, I'll leave that choice up to you.

```suggestion
/-- truncating the result of appending two bitvectors to the width of the right operand gives
exactly the right operand -/
theorem zeroExtend_append_eq {w v : Nat} (x : BitVec w) (y : BitVec v) :
    BitVec.zeroExtend v (x ++ y) = y := by
  ext
  simp only [truncate_append, Nat.le_refl, ↓reduceDIte, zeroExtend_eq]
```

alexkeizer (dismissed):
Thanks for the contribution! LGTM, but I've left some nits on how the theorem could be generalized further.

shigoel (approved):
Thanks, @luisacicolini!

PR #231, opened 2024-10-10
chore: Split out the finishing tactic aspect of simp_mem into mem_omega [2/?]

bollu:
This helps control how aggressive `simp_mem` is, since `simp_mem` now no longer closes memory goals, but rather expects the user to manually invoke `mem_omega` when necessary. Another impact of this change is that we split `mem_omega` into two tactics: `mem_omega` that does not expose `pairwiseSeparate` goals, and `mem_omega!`, which does. This also helps control the performance of `mem_omega`, and hopefully, this ensures that the user is careful before exposing a full `O(n^2)` set of constraints to the user. 

--- 

I was hoping that we would see a major performance difference. We do see some evidence for improvement in the `Experiments/MemoryAliasing.lean` (the numbers are extremely consistent across runs:

```
### Old timings
lake build Proofs.Experiments.MemoryAliasing  2.86s user 0.34s system 92% cpu 3.466 total

### New timings
lake build Proofs.Experiments.MemoryAliasing  2.76s user 0.33s system 92% cpu 3.332 total
```

---

However, on the much larger `Memcpy.lean`, these types of considerations seem to just not matter:

```
### Old timings
lake build Proofs.Experiments.Memcpy.MemCpyVCG  31.79s user 0.81s system 99% cpu 32.878 total

### New timings
lake build Proofs.Experiments.Memcpy.MemCpyVCG  32.79s user 0.80s system 99% cpu 33.870 total
```

This is a tad disappointing, but such is life. Onward to the next refactor. This is stacked on top of #230


Arm/Memory/MemOmega.lean
```diff
@@ -0,0 +1,134 @@
...
+/--
+Implement the `mem_omega` tactic, which unfolds information about memory
+in terms of
+-/
+syntax (name := mem_omega) "mem_omega" (Lean.Parser.Tactic.config)? : tactic
+
+/--
+Implement the `mem_omega` tactic frontend.
+-/
+syntax (name := mem_omega_bang) "mem_omega!" (Lean.Parser.Tactic.config)? : tactic
```
shigoel, line 118: Please document `omega!` and its difference to `omega`.

bollu: Done!

PR #241, opened 2024-10-16
feat: light-weight basic quotation macro

alexkeizer:
### Description:

I finally got annoyed enough at the lack of Qq, that I figured we can just roll our own basic version to get 90% of the benefit. So, I implemented a very basic `ql(..)` macro. Crucially, I only attempt to deal with closed terms, so we avoid a lot of the complexity that QQ has to deal with anti-quotations. Similarly, we avoid the performance problems of QQ because we don't attempt to deal with expression matching either.

To implement the macro, we had to take the `ToExpr Expr` instance from Mathlib, which meant pulling in the ToExpr derive handler (which isn't a bad thing, that derive handler will be generally useful). That is, `DeriveToExpr.lean`, `ToExpr.lean` and `ToLevel.lean` are taken from Mathlib, with only very slight modifications to eliminate any further Mathlib dependencies to do with pretty-printing.

Then, `Tactics/QuoteLight.lean` is actuallu new code. The implementation of the `ql(..)` macro is quite simple, only about 50 lines of meta-code. The rest of the file consists of a few basic tests.

Even with this limited scope, the macro will be useful to simplify a bunch of our meta-code, removing some of the tedium involved and improving readability. For example, we currently have `mkApp (.const ``List.nil [0]) (mkConst ``StateField)` somewhere in the code, with the macro we can simplify this to `ql(@List.nil StateField)`.

### Testing:

### License:

### DIFF
--- a/Tactics/QuoteLight.lean
+++ b/Tactics/QuoteLight.lean
@@ -1,0 +1,104 @@
+import Lean
+import Tactics.QuoteLight.ToExpr
+
+open Lean Meta Elab.Term
+
+namespace QuoteLight
+
+/-- `QL α` is a Lean `Expr` of type `α`, see the `ql(..)` and `QL(..)` macros
+for
+
+It's primary utility is to serve as machine-checked documentation of the
+expected type of expressions. However, the index is *not* actually checked,
+so it is entirely possible to construt  -/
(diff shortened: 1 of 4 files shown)


bollu:
LGTM, we paired on this, and I like the idea.

